Issue #9, opened 2020-09-21
Create new sidecar image to override C* version


grighetto:
While talking with John Sanda, he let me know that in order to use a custom C* image in the Cass-Operator, we have to create a new sidecar image first because the C* version is hardcoded in the Dockerfile, for example: https://github.com/datastax/management-api-for-apache-cassandra/blob/master/Dockerfile-3_11

┆Issue is synchronized with this [Jira Task](https://datastax-oss.atlassian.net/browse/AD-107) by [Unito](https://www.unito.io)
┆Issue Number: AD-107

Issue #32, opened 2020-10-19
Build images for all dependencies before running the workflow


grighetto:
Up until now, we've been building all the dependencies that have no official Docker images on the fly (Gemini, Spark, cassandra-diff) within the k8s cluster itself at every execution of Adelphi in order to have a consistent and simple experience locally or across cloud envs, but as we add more tools, the startup time adds up.

Instead, we should start by at least providing Dockerfiles such that one could build these images once and reuse them locally with k3d. Later, we could publish these images to a public registry.

┆Issue is synchronized with this [Jira Task](https://datastax-oss.atlassian.net/browse/AD-85) by [Unito](https://www.unito.io)
┆Issue Number: AD-85

grighetto:
If we want to guarantee a test run to be reproducible for a long time, we need the tools we use to be available. Things like cassandra-diff and Gemini have no official public images.
We should consider building and publishing them.

absurdfarce:
This seems like another argument for building "fatter" images which already contain the tools.  The image itself functions as a "checkpoint" for the combination of the specific tools in question.

An example: I'm imagining migrating the current "spark-image" template to an image which already has a precise version of Spark + cassandra-diff installed.  This image is itself versioned, which means we can explicitly say that a user running v1.2.3 of the image is using Spark 3.0.1 and cassandra-diff foo [1].  This change alone _greatly_ facilitates reproducibility and general uniformity.  It also makes everything run faster (under the assumption that the increase in download times for a larger image is more than offset by the removal of the need to download, build and install individual binaries as is done in the current workflow).

[1] Since cassandra-diff doesn't have explicit releases, or at least don't yet, this likely equates to a git hash

grighetto:
@absurdfarce Those are all good points, thanks!

I think I'd prefer individual images per tool instead of a fat image with everything for a couple of reasons:
1- some users may choose not to run all the available tools, so they could download just the images they need. That makes for smaller images and faster downloads.
2- provides isolation of dependencies - for example, some tools require newer versions of the JVM to run, so if they live in different images, we don't have to worry about clashes or incompatibilities.
I agree with the specific example of Spark and cassandra-diff though, because they to have to be in the same image (the cassandra-diff jars have to be in the Spark classpath).

What I'm envisioning is that the top-level version would be at the Helm chart level and its YAML files would point to (pin) specific versions of the tools' images.

PR #34, opened 2020-10-22
#9 Use a C* build from the source as the target cluster image

grighetto:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198513180644180/1198884156984741) by [Unito](https://www.unito.io/learn-more)


helm/adelphi/templates/cassandra-image-builder.yaml
```diff
@@ -0,0 +1,69 @@
...
+    - name: clone-cassandra
+      image: alpine/git
+      workingDir: /download
+      command:
+        - sh
+        - -c
+      args:
+        - git clone https://github.com/apache/cassandra.git cassandra;
+          cd cassandra;
+          git checkout {{`{{inputs.parameters.commit_hash}}`}};
```
absurdfarce, line 37: Nit: in the case of, say, "cassandra-4.0-beta2" (which looks to be the default given values.yaml) this represents a tag rather than a commit hash.  "tree-ish" is probably the right name here (since that's ultimately what's being passed to "git checkout") but that name... isn't great.  This name could be a branch, it could be a tag, or it could indeed be a commit.  Maybe... "git_identifier" or something?

That said, the term "tree-ish" is reasonably well understood, at least within the git community.  So maybe we just go with that... ?

I guess I lean a bit more towards "git_identifier"... but I could be swayed either way.

helm/adelphi/templates/cassandra-dockerfile-configmap.yaml
```diff
@@ -0,0 +1,207 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cassandra-dockerfile-configmap
+  namespace: {{ .Values.namespace }}
+data:
+  Dockerfile: |
```
absurdfarce, line 7: The remainder of this file represents an implicit import of the logic in https://github.com/docker-library/cassandra/blob/b6554329fe112243d16861b441067227eedcbdf9/4.0/Dockerfile and https://github.com/docker-library/cassandra/blob/c03e9828c699b8c22bcb17f82356ae90123d1d10/4.0/docker-entrypoint.sh into the project.  I'm not sure this is ideal for the following reasons:

* Fixes/enhancements to these files over time won't be automatically picked up and re-used here
* These files aren't easily identifiable in the context in which they serve, which makes them harder to find and review.  If we had a file named "Dockerfile" in a directory named "cassandra-4.0" the intent is pretty clear, but as it stands we have to get all of this from a config map.

Is there some reason we can't grab these files directly (from github, via git, or via some other mechanism) and then build the YAML programmatically to include them here?  I'm just concerned we're seeing ourselves up for a maintenance nightmare by bringing this stuff in in this way.

helm/adelphi/templates/workflow-adelphi.yaml
```diff
@@ -1,19 +1,44 @@
 apiVersion: argoproj.io/v1alpha1
 kind: Workflow
 metadata:
-  name: workflow-nosqlbench-cassdiff-{{ .Release.Revision }}
+  name: workflow-adelphi-{{ .Release.Revision }}
```
absurdfarce, line 4: 👍

helm/adelphi/templates/cassandra-image-builder.yaml
```diff
@@ -0,0 +1,69 @@
...
+    inputs:
+      parameters:
+      - name: registry_ip
+      - name: commit_hash
+    container:
+      image: gcr.io/kaniko-project/executor:v1.0.0
+      args:
+      - --dockerfile=/dockerfile/Dockerfile
+      - --context=dir:///workspace
+      - {{`--destination={{inputs.parameters.registry_ip}}:30000/cassandra-quality/cassandra`}}
```
absurdfarce, line 20: Don't we want this to be "adelphi" rather than "cassandra-quality"?

helm/adelphi/templates/cassandra-mgmt-image-builder.yaml
```diff
@@ -0,0 +1,72 @@
...
+  - name: cassandra-mgmt-image-template
+    inputs:
+      parameters:
+      - name: registry_ip
+    container:
+      image: gcr.io/kaniko-project/executor:v1.0.0
+      args:
+      - --dockerfile=/generated/Dockerfile
+      - --context=dir:///build
+      - {{`--destination={{inputs.parameters.registry_ip}}:30000/cassandra-quality/management-api-for-apache-cassandra`}}
```
absurdfarce, line 19: Same here: s/cassandra-quality/adelphi/

absurdfarce (commented):
Seems reasonable for what it does, but I have some serious concerns about bringing all this code into the code base rather than referencing external sources.  I think I have an even larger concern about the increase in work that we're now asking the k8s cluster to do; we're now adding logic to download, build and package C* into Docker containers.  This cost is paid _every time we run the Helm chart_ (unless there's a bypass or some kind of opt-out logic I missed).  I'm very concerned that this general approach just won't scale over time as more and more features are added.

helm/adelphi/templates/cassandra-dockerfile-configmap.yaml
grighetto: The Dockerfile in Adelphi is a modified version from the one you linked in docker-library. The original own downloads a prebuilt tarball from the Apache site, while ours copies it from the volume where we buid C* from the source.

Usually these files are "templates" that we modify to our needs, the DataStax Management API follows a similar approach: https://github.com/datastax/management-api-for-apache-cassandra/blob/master/upstream-4.0/Dockerfile

Adelphi is all about building images and orchestrating containers, so I think it makes sense to have those things under our control instead of having it change under our feet.
Also, that Dockerfile is pretty straightforward and looks stable, it hasn't changed much in docker-file since older versions of C*.

helm/adelphi/templates/cassandra-image-builder.yaml
grighetto: Fixed in the follow-up PR #35

helm/adelphi/templates/cassandra-mgmt-image-builder.yaml
grighetto: Fixed in the follow-up PR #35

helm/adelphi/templates/cassandra-dockerfile-configmap.yaml
grighetto: Another example of this approach are the CRDs manifests for C* Operator and Argo.
We can't import them as subcharts in Helm (because they're CRDs) and they are also tweaked for changing RoleBindings and stuff like that. The recommendation I heard from Jim Dickinson (from the C* Operator project) is that these files are mostly templates that have to be modified depending on one's needs.

grighetto:
> I have some serious concerns about bringing all this code into the code base rather than referencing external sources

I tried to address that concern in the comments above, but simply put the idea is that IMO Adelphi should own the image building process and we can't use the referenced resources directly because they do other things that we don't want to do. We have modified versions.

> I think I have an even larger concern about the increase in work that we're now asking the k8s cluster to do; we're now adding logic to download, build and package C* into Docker containers. This cost is paid every time we run the Helm chart 

For serious testing, most people will run this on some cloud environment. In my GKE test, it took 4 minutes to build C* from the source and 3 minutes to build the C* Management API, I think that's mostly negligible for workload generation tests that will run for maybe 30min on each anonymized schema (so it could be a few hours total in CI).

The image has to come from somewhere before running the workflow and it has to be built from the source since we'll want to test trunk or a dev branch. In a follow-up work we can let the user specify a pre-built tarball for local testing (let's investigate that in #32), but I still think there's value in building C* in the k8s cluster for real testing for the following reasons:
1- it guarantees reproducibility (we know what's inside the tarball and we'll have a git hash)
2- makes it easy to test with different JDK versions - Adelphi should be able to test C* compiled with JDK 8 and run with both JDK 8 and 11; and compiled and run with JDK 11.
3- the management of the custom image is transparent, otherwise each cloud vendor has its way to publish an image. Also, the image is destroy when the k8s cluster is destroyed, we don't have to worry about cleaning things up.
4- we make no assumptions about the user environment. We want this to be as simple as possible to execute, if everything is self-contained inside the workflow/k8s, we don't have to worry about the user's operating system, git, java version, ant, etc.

@absurdfarce Let me know if that clarifies things a bit or if you still have any questions or suggestions for improvement.

helm/adelphi/templates/cassandra-image-builder.yaml
grighetto: @absurdfarce I gave this some more thought as well, I wasn't really happy with "hash" either for the same reasons you mentioned.
Tree-ish is a bit confusing to me actually (it makes me think of a path) even though that's what I see in the "git checkout" command help. Strictly speaking, it seems like "commit" and "tree" are two different object types in git: https://matthew-brett.github.io/curious-git/git_object_types.html.

Having said that, I'm also leaning towards "git_identifier" or maybe even "git_checkout" to be self-explanatory (either way, this will be documented).
Let me know what you think.

grighetto:
@absurdfarce I just fixed a conflict with the base branch (it turns out I had also added that same `-DskipTests` option at the same line you did).
Please, let me know if you have any other feedback, otherwise we should merge this.

helm/adelphi/templates/cassandra-image-builder.yaml
absurdfarce: I do think "git_identifier" is the way to go here.  We're talking about something in the git log so "checkout" doesn't really seem right since that's not really what the user is specifying with that value.  That value specifies a tag or a specific commit, something like that... and "identifier" seems general enough to account for all of those options.

absurdfarce (approved):
After discussing this with @grighetto recently the plan is to move towards Docker images (where possible) for C* versions falling back to source builds only when necessary.  Approving this PR to get source builds in place; we'll add the logic around using the image (if available) later.

grighetto:
Thanks @absurdfarce.

For future reference, these are the follow-up improvement tickets:

https://github.com/datastax/adelphi/issues/32
https://github.com/datastax/adelphi/issues/39

PR #42, opened 2020-10-30
27 integrate schema with gemini

absurdfarce:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198513180644180/1198984111002425) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/helm/adelphi/templates/gemini-job.yaml
+++ b/helm/adelphi/templates/gemini-job.yaml
@@ -49,7 +49,7 @@ spec:
       - name: workspace
         mountPath: /workspace
     - name: extract-schema
-      image: frolvlad/alpine-python2
+      image: python:3.10.0a1-alpine3.12
       command: ["sh", "-c", "-e"]
       args:
         - echo "Extract Gemini schema using anonymizer script";
--- a/helm/adelphi/templates/gemini-job.yaml
+++ b/helm/adelphi/templates/gemini-job.yaml
@@ -22,6 +22,7 @@ spec:
           --test-cluster={{ .Values.target.clusterName }}-{{ .Values.target.dc }}-service
           --test-username={{ .Values.target.clusterName }}-superuser
           --test-password="`cat /target-secret/password`"
+          --schema=/workspace/gemini_schema.json
           --duration=10m
           --fail-fast
       dnsConfig:
--- a/helm/adelphi/templates/gemini-job.yaml
+++ b/helm/adelphi/templates/gemini-job.yaml
@@ -48,6 +48,21 @@ spec:
       volumeMounts:
       - name: workspace
         mountPath: /workspace
+    - name: extract-schema
+      image: frolvlad/alpine-python2
+      command: ["sh", "-c", "-e"]
+      args:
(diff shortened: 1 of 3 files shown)


helm/adelphi/templates/schema-job.yaml
```diff
@@ -9,7 +9,7 @@ spec:
   templates:
   - name: configure-schema
     container:
-      image: alpine
+      image: frolvlad/alpine-python2
```
absurdfarce, line 12: Base alpine image didn't include Python

helm/adelphi/templates/schema-job.yaml
```diff
@@ -20,6 +20,9 @@ spec:
           echo "Configuring schema on the target cluster...";
           cat /schema/schema.cql | kubectl exec -n {{ .Values.namespace }} -i -t -c cassandra {{ .Values.target.clusterName }}-{{ .Values.target.dc }}-default-sts-0 -- /opt/cassandra/bin/cqlsh -u {{ .Values.target.clusterName }}-superuser -p $(cat /secret-target/password);
           echo "Done";
+          pip install cassandra-driver;
+          wget -q https://raw.githubusercontent.com/datastax/adelphi/master/schema-anonymizer/anonymizer.py;
```
absurdfarce, line 24: This setup will also be removed once most of this installation logic is converted to the Docker images

helm/adelphi/values.yaml
```diff
@@ -58,6 +58,7 @@ cql_schema: |-
     value text,
     PRIMARY KEY (key)
   );
+gemini_schema_keyspace: testkeyspace
```
absurdfarce, line 61: Anonymizer script only supports a single keyspace when generating Gemini output.  This val is used to constrain that operation.

helm/adelphi/templates/schema-job.yaml
```diff
@@ -45,4 +45,4 @@ spec:
         secretName: {{ .Values.target.clusterName }}-superuser
         items:
         - key: password
-          path: password
\ No newline at end of file
+          path: password
```
absurdfarce, line 48: A side effect of an earlier impl which tried to perform schema extraction in the schema-job template

absurdfarce:
Observed clean run of entire Argo workflow (including gemini-oracle-sut template) with this configuration

helm/adelphi/templates/gemini-job.yaml
```diff
@@ -47,6 +48,21 @@ spec:
       volumeMounts:
       - name: workspace
         mountPath: /workspace
+    - name: extract-schema
+      image: python:3.10.0a1-alpine3.12
```
grighetto, line 52: @absurdfarce fyi, I replaced the original image you had here with the official one for Python.

grighetto:
@absurdfarce This works fine 👍 but I noticed it anonymizes the schema again when it runs the script to extract the Gemini schema. I think that's not desirable since we want to test the original schemas as they have been contributed to the project.
I'll open a bug ticket to fix that separately, it doesn't need to block this PR.

grighetto (approved):

Issue #20, opened 2020-09-29
Report test results


grighetto:
At the moment, we can simply tell whether the tests worked by having a succeeded workflow execution, but we should collect the results generated by cassandra-diff, Gemini and NoSQLBench and report them to the user.

* cassandra-diff results are stored in the cassandra-diff keyspace of the target cluster
* Gemini results are printed to the console in the pod it executed

┆Issue is synchronized with this [Jira Task](https://datastax-oss.atlassian.net/browse/AD-75) by [Unito](https://www.unito.io)
┆Issue Number: AD-75

grighetto:
Successfully tested on GKE. Resolving issue.

<img width="1415" alt="gke-collect-results-success" src="https://user-images.githubusercontent.com/413792/98197593-d98b1380-1f05-11eb-91ee-b73426eb9267.png">
<img width="724" alt="browse-result-files" src="https://user-images.githubusercontent.com/413792/98197610-df80f480-1f05-11eb-872d-5eb5012c9750.png">

Issue #39, opened 2020-10-27
Use a public C* image by default if git hash is not specified


grighetto:
Issue #9 introduced a way to build the C* image from the source to use it in the target cluster and made it the default behavior, but instead we should give preference to using a publicly available image from Dockerhub, if possible.

The goal of this ticket is to add some conditional in the Argo workflow that checks the `git_hash` configuration and, if not provided, skips building the image.

┆Issue is synchronized with this [Jira Task](https://datastax-oss.atlassian.net/browse/AD-89) by [Unito](https://www.unito.io)
┆Issue Number: AD-89

grighetto:
This is included in #44

PR #44, opened 2020-11-03
#20 Collect test results

grighetto:
This PR aims at collecting the test results from Gemini, cassandra-diff and NoSQLBench at a single place that can be easily exposed through a web server (nginx).

Note: GKE doesn't support ReadWriteMany access mode for volumes at the moment and that prevented us from writing the results from different nodes simultaneously. One way to overcome that limitation would be to use S3 or some other dedicated storage, but that would require more operational work. Instead, we followed this approach: https://github.com/kubernetes/examples/tree/master/staging/volumes/nfs. Having an NFS server allows for simultaneous writes while keeping things simple and self-contained. We can revisit this if it ever becomes a scaling issue.

In order to access the web page locally, run: `kubectl -n cass-operator port-forward results-server 8080:8080` then hit http://localhost:8080 and browse the result files.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198513180644180/1199090435845041) by [Unito](https://www.unito.io/learn-more)


helm/adelphi/templates/cassandra-clusters.yaml
```diff
@@ -77,7 +77,9 @@ spec:
           clusterName: {{ .Values.target.clusterName }}
           serverType: cassandra
           serverVersion: {{ .Values.target.version }}
+          {{- if .Values.target.commit_hash }}
```
grighetto, line 80: This is for #39

helm/adelphi/templates/workflow-adelphi.yaml
```diff
@@ -15,6 +15,7 @@ spec:
           template: registry-template
       - name: build-cassandra
         depends: "start-registry"
+        when: "{{ default false .Values.target.commit_hash }}"
```
grighetto, line 18: For #39 - if commit_hash is not specified, it will skip this step

helm/adelphi/templates/workflow-adelphi.yaml
```diff
@@ -26,6 +27,7 @@ spec:
           template: cassandra-image-template
       - name: build-cassandra-mgmt-api
         depends: "build-cassandra"
+        when: "{{ default false .Values.target.commit_hash }}"
```
grighetto, line 30: For #39 - if commit_hash is not specified, it will skip this step

helm/adelphi/templates/results-server.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: results-server
+  namespace: {{ .Values.namespace }}
+spec:
+  containers:
+    - name: results
+      image: nginx
```
grighetto, line 9: web server to expose the result files for inspection or download

helm/adelphi/templates/nfs-server.yaml
```diff
@@ -0,0 +1,33 @@
...
+  replicas: 1
+  selector:
+    role: nfs-server
+  template:
+    metadata:
+      labels:
+        role: nfs-server
+    spec:
+      containers:
+      - name: nfs-server
```
grighetto, line 16: this server is here to overcome the ReadWriteMany limitation in GKE, while keeping the workflow self-contained and not depending on extra operational work

helm/adelphi/templates/cassandra-clusters.yaml
absurdfarce: How does this work exactly?  If a commit hash _isn't_ specified... how do we tell what value is used for "serverImage" here?  Is it just empty?

helm/adelphi/templates/workflow-adelphi.yaml
absurdfarce, line 30: We have a single value "commit_hash" to represent both the C* tag/commit/whatever and the similar entity for the management API.  Is it valid that both C* and the mgmt API will use the same tag names... or do we need to allow users to specify distinct "commit hash" values, one for C* and one for the management API?

helm/adelphi/templates/cassandra-clusters.yaml
```diff
@@ -77,7 +77,9 @@ spec:
...
           serverImage: {{`{{inputs.parameters.registry_ip}}`}}:30000/adelphi/management-api-for-apache-cassandra
```
absurdfarce, line 81: Is this... right?  Shouldn't this be the C* image and not the management API image?

helm/adelphi/templates/nosqlbench-jobs.yaml
```diff
@@ -12,6 +12,7 @@ spec:
       image: nosqlbench/nosqlbench:latest
       args: [
         "--show-stacktraces",
+        "--report-csv-to=/results/nosqlbench-source",
```
absurdfarce, line 15: "nosqlbench-source" is... actually the CSV file itself and not a directory?  I ask only because the Gemini template explicitly did a "mkdir" before running the app.  If this is intended to be the CSV file itself then clearly making a directory is unnecessary... just wanted to understand.

helm/adelphi/templates/nosqlbench-jobs.yaml
```diff
@@ -38,11 +41,15 @@ spec:
...
         name: nosqlbench-workload-configmap
+    - name: results-pv
+      persistentVolumeClaim:
+        claimName: results-pvc
   - name: nosqlbench-target
     container:
       image: nosqlbench/nosqlbench:latest
       args: [
         "--show-stacktraces",
+        "--report-csv-to=/results/nosqlbench-target",
```
absurdfarce, line 52: Same question re: is this just a CSV file (see above for more detail)

absurdfarce:
I've looked over the mods to the existing templates + the new results server stuff and those look pretty much as expected.  I have a few questions at various points but I don't think there's anything too big in there.

I don't understand the persistent volume/NFS stuff well enough to review it yet; I certainly wasn't expecting to see NFS pop up as part of the resolution of this issue. :)  I need to review the info @grighetto passed along on this point to understand more completely what's going on.  Plan is to do that and pick this back up tomorrow... er, later today (Wednesday, that is).

helm/adelphi/templates/cassandra-clusters.yaml
grighetto: @absurdfarce take a look at https://docs.datastax.com/en/cass-operator/doc/cass-operator/cassOperatorProvisioning.html#Specifytheservertypeandversion.

If the user specifies a commit_hash in Adelphi, it means he wants to use an image built from the source (say, from trunk or a dev branch), in that case Adelphi will build the image and set `serverImage` accordingly. 
Otherwise, they'll just set the `version` to some C* release version and the operator will know what to do.

It sounds like you're referring to a corner case where they specify a git hash and by chance that matches an existing C* release. That's indeed not covered here, it doesn't seem like that would happen to often though, we should advise them to use a version number instead.

grighetto:
@absurdfarce It sounds like most of your questions gravitate towards the usage of a custom C* image with the Operator, I think this can help clarifying things a bit: https://docs.datastax.com/en/cass-operator/doc/cass-operator/cassOperatorProvisioning.html#Specifytheservertypeandversion.
Check the "Specify the server type and version" section.

```
Using a specific Cassandra image - for example:

apiVersion: cassandra.datastax.com/v1beta1
kind: CassandraDatacenter
metadata:
  name: dtcntr
spec:
  serverType: cassandra
  serverVersion: 3.11.7
  serverImage: my-private-docker-registry.example.com/datastax/cassandra-mgmtapi-3_11_7
```

So it works like this:
1- Build a custom C* image from a git hash.
2- Build a custom C* mgmt API image that pulls the image from step 1.
3- Set the image from step 2 in the `serverImage` property of the Operator config.

helm/adelphi/templates/workflow-adelphi.yaml
grighetto: I'm not sure I understand your question here, I think there's some confusion on how the Operator works. Please, check the details I commented below and let me know if this is still a valid concern.

helm/adelphi/templates/cassandra-clusters.yaml
grighetto: Yes, it's correct, this has to be the MGMT API image. If you try it on k3d or GKE you'll see the C* cluster will be provisioned with the correct git hash version.

helm/adelphi/templates/nosqlbench-jobs.yaml
grighetto: NoSQLBench writes a bunch of output files into that folder.

helm/adelphi/templates/nosqlbench-jobs.yaml
grighetto: No, it's a folder.

grighetto:
>  I certainly wasn't expecting to see NFS pop up as part of the resolution of this issue. :)

Me neither :)
But it's the only solution I can think of right now for accommodating different scenarios without extra operational work: people running locally with k3d, running on GKE or EKS.
But I'm open to suggestions here as always.

In the future we can probably write the test results to a db, maybe have a third C* cluster that stores the results, but that's beyond the scope of this work.

grighetto:
gke-collect-results-success
browse-result-files

PR successfully tested on GKE.
Results are collected into a single place that I can browse through after the workflow is finished.
Better presentation and Graphite integration is coming in future work.

@absurdfarce I'll go ahead and merge this PR in the interest of time, it sounded like you didn't have any major concerns, feel free to take another pass and we can follow up in another PR if needed.

absurdfarce (approved):

PR #47, opened 2020-11-11
43 too much anonymizing

absurdfarce:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198513180644180/1198683676223747) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/schema-anonymizer/anonymizer.py
+++ b/schema-anonymizer/anonymizer.py
@@ -138,7 +138,10 @@ def get_name(original_name, prefix):
     """
     Looks up the anonymized name for the provided original name in the cache.
     If not present, one is created, inserted into the cache and returned.
+    If schema anonymization is not requested simply return the  originl name.
     """
+    if not anonymize:
+        return original_name
     count = len(name_map[prefix])
     anonymized_named_prefixed = "%s_%s" % (prefix, count)
     return name_map[prefix].setdefault(original_name, anonymized_named_prefixed)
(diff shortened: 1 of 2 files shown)


schema-anonymizer/anonymizer.py
```diff
@@ -273,6 +276,8 @@ def transform_if_not_exists(cql_string):
                     help='Replication factor to override original setting. Optional.')
 parser.add_argument('--format', metavar='cql', default='cql', choices=['cql', 'gemini'],
                     help='Output format for the schema. cql or gemini (json). Default is cql.')
+parser.add_argument('--no-anonymize', dest='anonymize', action='store_false', help='Disable anonymization functions')
```
absurdfarce, line 279: This argument was deliberately selected so that anonymizing of schemas is still the default process.  In order to preserve confidentiality of submitted schemas we should anonymize in all cases _except_ when a user has explicitly asked for this not to occur.

grighetto:
Verified the script in standalone manner and also as part of the workflow in GKE, works as expected. 👍 
If the `--no-anonymize` is not passed, it works as before.

Below is the output of the Gemini run that shows the names remain the same as given:

```
{
     "gemini_version": "1.7.3",
     "result": {
         "write_ops": 57687,
         "write_errors": 0,
         "read_ops": 1277,
         "read_errors": 0
     },
     "schema": {
         "keyspace": {
             "name": "testkeyspace",
             "replication": {
                 "class": "SimpleStrategy",
                 "replication_factor": "1"
             },
             "oracle_replication": {
                 "class": "SimpleStrategy",
                 "replication_factor": "1"
             }
         },
         "tables": [
             {
                 "name": "testtable",
                 "partition_keys": [
                     {
                         "name": "key",
                         "type": "text"
                     }
                 ],
                 "clustering_keys": [],
                 "columns": [
                     {
                         "name": "value",
                         "type": "text"
                     }
                 ],
                 "known_issues": null
             }
         ]
     }
 }
```

grighetto (approved):
+1

PR #49, opened 2020-11-13
Preparation for GH schema contrib support

absurdfarce:
A lot of refactoring, abstracting away some of the existing functionality so that bits of it can be re-used for schema contribution

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198513180644180/1198832404957914) by [Unito](https://www.unito.io/learn-more)


python/adelphi/adelphi/__init__.py
```diff
@@ -0,0 +1,10 @@
+# list of system keyspaces to filter out
+system_keyspaces = ["system",
+                    "system_schema",
+                    "system_traces",
+                    "system_auth",
+                    "system_distributed",
+                    "system_virtual_schema",
+                    "system_views"]
```
absurdfarce, line 8: I'm not really sure this is the right place for this logic... but I couldn't find another natural place to put it.  Perhaps it should go into store.py since that's where it's actually used... that might make the most sense.

python/adelphi/adelphi/anonymize.py
```diff
@@ -0,0 +1,96 @@
...
+name_map = {
+    KEYSPACE_PREFIX: {},
+    TABLE_PREFIX: {},
+    PARTITION_KEY_PREFIX: {},
+    CLUSTERING_KEY_PREFIX: {},
+    COLUMN_PREFIX: {},
+    TYPE_PREFIX: {},
+    FIELD_PREFIX: {},
+    INDEX_PREFIX: {}
+}
```
absurdfarce, line 25: Might need to re-think this: if we run multiple schemas through the anonymizer this probably should become a per-schema map which means we'd need to regenerate it on a per-keyspace basis.  It works as-is but the isolation offered by moving this to a per-keyspace thing might be (slightly) better.

python/adelphi/adelphi/store.py
```diff
@@ -0,0 +1,61 @@
+# Functions to facilitate interactions with the underlying data store
```
absurdfarce, line 1: Not using the name "cassandra" for this module in order to avoid any interference with the Cassandra driver

python/adelphi/bin/adelphi
```diff
@@ -0,0 +1,83 @@
...
+
+from adelphi import system_keyspaces
+from adelphi.cql import export_cql_schema
+from adelphi.gemini import export_gemini_schema, to_string
+from adelphi.store import with_cluster, filter_keyspaces_for_export
+
+logging.basicConfig(level=logging.INFO)
+log = logging.getLogger('adelphi')
+
+@click.group()
```
absurdfarce, line 17: See https://click.palletsprojects.com/en/7.x/ for more details

python/adelphi/adelphi/__init__.py
absurdfarce: Subsequently moved into adelphi.store

absurdfarce:
Note on use: args come _before_ the command is specified, so you're looking at something like:

> adelphi --hosts=192.168.1.1 --port=9142 --keyspaces=foo,bar export-cql

or

> adelphi --keyspaces=foo,bar export-gemini

python/adelphi/adelphi/anonymize.py
grighetto: Does it matter to have different anonymized names per schema? I think as long it's anonymous and consistent across the generated CQL statements, we don't need to worry about re-using a name.

python/adelphi/setup.py
```diff
@@ -0,0 +1,16 @@
...
+   author='Gianluca Righetto',
+   author_email='gianluca.righetto@datastax.com',
+   packages=['adelphi'],
+   scripts=['bin/adelphi'],
+   #url='http://pypi.python.org/pypi/PackageName/',
+   url='https://github.com/datastax/adelphi',
+   #license='LICENSE.txt',
+   description='Schema anonymizer for Adelphi',
+   #long_description=open('README.txt').read(),
+   install_requires=['cassandra-driver', 'click'],
```
grighetto, line 15: it's probably a good idea to pin the cassandra-driver to a specific version such that it doesn't update under our feet as it could export more schema information in future versions

python/adelphi/adelphi/anonymize.py
```diff
@@ -0,0 +1,96 @@
...
+    return name_map[prefix].setdefault(original_name, anonymized_named_prefixed)
+
+
+def anonymize_keyspace(keyspace):
+    keyspace.name = get_name(keyspace.name, KEYSPACE_PREFIX)
+
+    for table in keyspace.tables.values():
+        anonymize_table(table)
+
+    for table in keyspace.user_types.values():
```
grighetto, line 44: nit: this should be `for udt` (noticed this in my branch too)

grighetto (changes requested):
as for changes, besides pinning the driver version, I'm also seeing this when I try running the adelphi script:

```
/opt/anaconda3/lib/python3.8/site-packages/adelphi-0.1.0-py3.8.egg/EGG-INFO/scripts/adelphi", line 56
    print export_fn(export_keyspaces, schema, options)
          ^
SyntaxError: invalid syntax
```

python/adelphi/adelphi/anonymize.py
absurdfarce: Yeah, that's what I meant by "works as-is" above.  I'm not suggesting it's broken, I'm just wondering if it isn't more straightforward to manage on a per-keyspace basis.  Certainly not essential, just something I was musing about.

python/adelphi/setup.py
absurdfarce: Agreed, I'll fix this.

python/adelphi/adelphi/anonymize.py
absurdfarce: Ah, good catch!  I'll fix this as well.

absurdfarce:
Fixes for driver pin + UDT anonymization pushed.  @grighetto would you mind re-testing your "invalid syntax" case above?  I also noticed (and fixed) a dumb mistake I made in an earlier change and I'm wondering if that was the root cause of the behaviour you were seeing.

grighetto:
@absurdfarce I'm still seeing that error, I believe it's because `print` is a function in Python 3, it needs parentheses to be a function call.

grighetto:
Cool, we're past that error! Now I'm seeing this when running with default values (`adelphi export-cql`):

```
AttributeError: 'NoneType' object has no attribute 'replication_factor_info'
```

python/adelphi/adelphi/store.py
```diff
@@ -0,0 +1,87 @@
...
+def with_cluster(cluster_fn, hosts, port, username = None, password = None):
+    ep = ExecutionProfile(load_balancing_policy=default_lbp_factory())
+    cluster = Cluster(hosts, port=port, auth_provider=build_auth_provider(username,password), execution_profiles={EXEC_PROFILE_DEFAULT: ep})
+    log.info("Connecting to the cluster to get metadata...")
+    cluster.connect()
+    cluster_fn(cluster)
+    cluster.shutdown()
+
+
+def build_keyspace_objects(keyspaces, metadata):
```
absurdfarce, line 43: This seemed to be a better name for this function.  This guy isn't just about filtering keyspaces... he's also about getting a sequence of keyspace _objects_ for a given set of keyspace _names_.

python/adelphi/adelphi/store.py
```diff
@@ -0,0 +1,87 @@
...
+    all_keyspace_objs = [metadata.keyspaces[ks] for ks in keyspaces] if keyspaces is not None else metadata.keyspaces.values()
+
+    # Borrowed from itertools
+    def partition(pred, iterable):
+        t1, t2 = tee(iterable)
+        return list(filterfalse(pred, t1)), list(filter(pred, t2))
+
+    (failed,passed) = partition(lambda ks: ks.name not in system_keyspaces,all_keyspace_objs)
+    if failed:
+        log.info("Excluding system keyspaces " + ",".join((ks.name for ks in failed)))
```
absurdfarce, line 54: It seemed worthwhile to be _very_ explicit here.  It's especially true if we're modifying a list of keyspaces explicitly supplied by the user but even in the "include all keyspaces" case it's worth being explicit here.

python/adelphi/adelphi/store.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def set_replication_factor(selected_keyspaces, factor):
+    if not factor:
+        log.debug("No replication factor provided")
+    else:
+        for ks in selected_keyspaces:
+            if ks.virtual:
+                log.debug("Keyspace " + ks.name + " is a virtual keyspace")
+                continue
```
absurdfarce, line 84: Virtual keyspaces have a replication strategy of None so the modifications below will obviously lead to bad results in that case.

grighetto (approved):
LGTM 👍

PR #51, opened 2020-11-17
40 Github support

absurdfarce:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198513180644180/1198728282212511) by [Unito](https://www.unito.io/learn-more)


python/adelphi/bin/adelphi
```diff
@@ -2,39 +2,60 @@
...
 import logging
+import os
+import os.path
 from functools import partial
 
+try:
+    from tempfile import TemporaryDirectory
+except ImportError:
+    from backports.tempfile import TemporaryDirectory
+
```
absurdfarce, line 13: Context mgr version of temp management not available on py2k so we swap in a backport in that situation

python/adelphi/adelphi/gh.py
```diff
@@ -0,0 +1,34 @@
+# Functionality necessary to create a feature branch + pull request on Github
```
absurdfarce, line 1: We must avoid the name "github" for this module since it's use causes Python 2.x to have problems with the PyGithub import ("from github import ...") op that happens a few lines below.  This isn't a problem on Python 3.x, but to avoid the issue entirely I just renamed the module.

python/adelphi/bin/adelphi
```diff
@@ -2,39 +2,60 @@
...
 @click.option('--port', type=int, default=9042, help='Database RPC port. Default: 9042')
-# not implemented yet
-#@click.option('--output', metavar='/file/path', help='Output file. If not specified, it will write to stdout')
 @click.option('--username', help='Database username')
 @click.option('--password', help='Database password')
 @click.option('--keyspaces', help='Comma-separated list of keyspaces to include. If not specified, all keypaces will '
                          'be included, except system keypaces')
 @click.option('--rf', type=int, help='Replication factor to override original setting. Optional.')
 @click.option('--anonymize/--no-anonymize', default=True, help="Enable/disable schema anonymization")
+@click.option('--output-dir', help='Directory schema files should be written to. If not specified, it will write to stdout')
```
absurdfarce, line 33: A happy side effect: since we needed "export schemas to file" functionality for GH support anyway we get this functionality for free

python/adelphi/bin/adelphi
```diff
@@ -79,9 +107,27 @@ def export_gemini(ctx):
...
     with_cluster(my_fn, **({k:ctx.obj[k] for k in ["hosts","port","username","password"]}))
 
 
+@export.command()
+@click.option('--repo-user', help='Username for the repo used for schema storage on Github')
+@click.option('--repo-token', help='Access token for schema storage on Github')
+@click.option('--repo-name', help='Name of the repo used for schema storage on Github')
+@click.option('--author-name', help='Github username to be used for schema commits')
+@click.pass_context
+def github(ctx, repo_user, repo_token, repo_name, author_name):
```
absurdfarce, line 120: At the moment Adelphi requires us to pass in each of these args every time, which is just a little silly.  It also makes the invocation fairly large; a sample:

```
adelphi --no-anonymize --keyspaces=testkeyspace,spacecraft --output-dir=foo github --repo-user=absurdfarce --repo-token=<long repo token> --repo-name=schema-playground --author-name=grighetto
```

Click does support the ability to pull these args from environment variables (see https://click.palletsprojects.com/en/7.x/arguments/#environment-variables) and this is probably preferred for the Github-related configs since they won't change much from one run to another.

python/adelphi/setup.py
```diff
@@ -1,5 +1,16 @@
...
+
+dependencies = [
+    'cassandra-driver >= 3.24.0',
+    'click >= 7.1.2',
+    'PyGithub >= 1.45'
+    ]
+
+if not PY3:
+    dependencies.append('backports.tempfile')
+
```
absurdfarce, line 13: Add the backport of py3k tempfile functionality for Python 2.7 support.  Logic here borrowed from what the Python driver does.

python/adelphi/setup.py
```diff
@@ -1,5 +1,16 @@
 from setuptools import setup
 
+PY3 = sys.version_info[0] == 3
+
+dependencies = [
+    'cassandra-driver >= 3.24.0',
+    'click >= 7.1.2',
+    'PyGithub >= 1.45'
+    ]
```
absurdfarce, line 9: The versions specified here are the min versions currently used for Python 2.7.  The ">=" specifier allows Py3k to use larger versions as appropriate.

python/adelphi/bin/adelphi
```diff
@@ -16,55 +39,81 @@ log = logging.getLogger('adelphi')
...
+    output_dir = ctx.obj["output-dir"]
+
     metadata = cluster.metadata
     keyspace_objects = build_keyspace_objects(keyspace_names, metadata)
+
+    if ctx.obj['anonymize']:
+        # anonymize keyspace and its children
+        for keyspace_object in keyspace_objects:
+            anonymize_keyspace(keyspace_object)
+
```
absurdfarce, line 95: The introduction of output to files meant that we had to do anonymization earlier in the process.  Before it was being done in the CQL/Gemini export op which happens much further down the line.  We want it to happen earlier (and in this script) so that all subsequent log statements/file names/etc. will use anonymized names.

python/adelphi/bin/adelphi
```diff
@@ -79,9 +128,32 @@ def export_gemini(ctx):
...
+@click.option('--token', help='Access token for schema storage on Github')
+# TODO: Change the default to something real when we have something real
+@click.option('--repo', default="absurdfarce/schema-playground", help='Name of the repo used for schema storage on Github')
+@click.option('--author', help='Github username to be used for schema commits')
+@click.pass_context
+def github(ctx, token, repo, author):
+
+    if ctx.obj["anonymize"]:
+        log.error("All schemas contributed to Adelphi must be anonymized")
+        exit(6)
```
absurdfarce, line 145: We want to explicitly reject the contribution of non-anonymized schemas to the repo

python/adelphi/setup.py
```diff
@@ -1,5 +1,16 @@
 from setuptools import setup
 
+PY3 = sys.version_info[0] == 3
+
+dependencies = [
+    'cassandra-driver >= 3.24.0',
```
grighetto, line 6: I believe this should remain a static/pinned version instead of a lower bound.
We don't want to suddenly pick up a newer version that exposes more metadata that's not covered by the anonimyzer.

python/adelphi/adelphi/gh.py
```diff
@@ -0,0 +1,61 @@
...
+
+import logging
+import os
+import os.path
+import uuid
+
+logging.basicConfig(level=logging.INFO)
+log = logging.getLogger('adelphi')
+
+# We're assuming any storage repo will be created after the converstion to "main"
```
grighetto, line 33: comment typo: conversion

python/adelphi/bin/adelphi
```diff
@@ -16,72 +39,119 @@ log = logging.getLogger('adelphi')
...
+        exit(2)
+
+    # My kingdom for decent function composition :(
+    export_keyspaces(ctx, lambda x,y: format_gemini_schema(export_gemini_schema(x,y)), keyspace_objs)
+
+
+@export.command()
+@click.option('--token', help='Access token for schema storage on Github')
+@click.pass_context
+def github(ctx, token):
```
grighetto, line 135: wdyt about making the command name `submit`? we may move away from github for the schemas repo in the future, but the command would remain the same

absurdfarce: Funny you should ask about this... I was mulling this very question the other day. :)

"github" is a terrible name, so that definitely shouldn't be the answer long-term, as it binds us to a specific type of repo and/or submission process.  I finally settled on the name "contribute" as being the most generic.  What do you think of that?  I could live with "submit" I suppose but I really like the explicit suggestion of submitting to the project.

Whaddya think?

python/adelphi/adelphi/gh.py
absurdfarce: :facepalm: Will fix in the next push

python/adelphi/setup.py
absurdfarce: I think this definitely could use another pass.

The original intent here was to pin deps to the lowest supported version for Python 2.7 while allowing them to go higher for Py3k.  My thinking was that projects would have a final version that ran on Python 2.7; PyGithub certainly does this.  So I wanted to make sure that baseline was supported while allowing Py3k instances to use higher versions.

This was my original attempt at dep management; it came before the subsequent logic which introduces an explicit check for Py3k + changes to deps in case it finds it.  That would probably be a cleaner way to handle this now.

Interestingly, it's also worth noting that PyGithub might be an exception.  The cassandra-driver package works with both 2.7 and 3.x so it could be explicitly pinned.  The same is true for click; both 2.7 and 3.x use 7.1.2.

There's _also_ the point that even if new metadata is introduced that's actually okay as long as they maintain backwards compatibility with the current version.  The better way to express _that_ idea would be to use something like Ruby's ~> comparison (at least as long as the package supports semantic versioning).  Moving to something like that may be worthwhile too.

So yeah, in short... I wanna take another look at this. :(

python/adelphi/setup.py
absurdfarce: Updated: shifted to compatible version specifiers as discussed in PEP-440

python/adelphi/bin/adelphi
grighetto: "contribute" sounds good and precise to me 👍

grighetto (approved):
Only nit was the cassandra-driver version that should be pinned to a specific version, other than that, 👍

Issue #41, opened 2020-10-30
Rename commit_hash param to git_identifier


grighetto:
Per discussion in https://github.com/datastax/adelphi/pull/34#discussion_r513624454

┆Issue is synchronized with this [Jira Task](https://datastax-oss.atlassian.net/browse/AD-92) by [Unito](https://www.unito.io)
┆Issue Number: AD-92

grighetto:
Pulling this lhf into the current milestone since I'm touching a related area of the code.

PR #54, opened 2020-11-23
#3 UDT support / Overall Unit & Integration tests

grighetto:
This PR adds support for UDT anonymization and includes unit and integration tests for all anonymized elements.

For the integration tests, there's a base CQL schema (generated programmatically) that is executed against a few C* versions in Docker, then the anonymizer is executed and the output is compared to the expected CQL file in the "schemas" folder.

The unit tests are mostly self-explanatory and they validate a few complex UDTs as per the original issue #3.

Steps to run the tests:
- Build the Python package as usual with: `python3 setup.py install`
- From `python/adelphi/tests/unit`, run the unit tests with: `pytest`
- From `python/adelphi/integration-tests`, run the integration tests with: `./adelphi-docker-tests.sh`


.gitignore
```diff
@@ -1,3 +1,6 @@
+# Test output files
+output/
```
grighetto, line 2: the integration tests output go there

helm/adelphi/templates/cassandra-clusters.yaml
```diff
@@ -77,7 +77,7 @@ spec:
           clusterName: {{ .Values.target.clusterName }}
           serverType: cassandra
           serverVersion: {{ .Values.target.version }}
-          {{- if .Values.target.commit_hash }}
+          {{- if .Values.target.git_identifier }}
```
grighetto, line 80: this is for #41

python/adelphi/adelphi/anonymize.py
```diff
@@ -1,12 +1,10 @@
+import re
 # Functions and constants related to the anonymization process
-
 from adelphi.store import get_standard_columns_from_table_metadata
 
 # default prefixes for the anonymized names
 KEYSPACE_PREFIX = "ks"
 TABLE_PREFIX = "tbl"
-PARTITION_KEY_PREFIX = "pk"
-CLUSTERING_KEY_PREFIX = "ck"
```
grighetto, line 9: Splitting pk/ck from regular columns added some unneeded complexity (some names might be clustering keys in one schema but not in another). The original requirement was to anonymize all columns to `col_`, so I converged to that.

python/adelphi/adelphi/anonymize.py
```diff
@@ -38,59 +33,70 @@ def get_name(original_name, prefix):
...
-        anonymize_udt(udt)
+    # remove functions, aggregates and views for now
+    keyspace.functions = {}
+    keyspace.aggregates = {}
+    keyspace.views = {}
 
+def anonymize_type(original_type):
+    name = original_type
+    for udt in name_map[TYPE_PREFIX]:
+        name = name.replace(udt, name_map[TYPE_PREFIX][udt])
```
grighetto, line 52: What this does is replace all occurrences of an UDT name by its anonymized name (it applies to column types and fields types/nested UDTs).

python/adelphi/adelphi/anonymize.py
```diff
@@ -38,59 +33,70 @@ def get_name(original_name, prefix):
...
+    # The CQL python driver holds the same object references for clustering keys
+    # and regular columns, so we have to separate them otherwise the names
+    # will be re-anonymized.
     for column in get_standard_columns_from_table_metadata(table):
-        anonymize_column(column, COLUMN_PREFIX)
+        anonymize_column(column)
 
     for index in list(table.indexes.values()):
-        if (index.index_options["target"] not in name_map[COLUMN_PREFIX].keys() and
-                index.index_options["target"] not in name_map[CLUSTERING_KEY_PREFIX].keys()):
```
grighetto, line 93: This smelled like pulling some C* validation into the anonymizer.

python/adelphi/adelphi/anonymize.py
```diff
@@ -38,59 +33,70 @@ def get_name(original_name, prefix):
...
+        if (index.kind == "CUSTOM"):
```
grighetto, line 99: We want to skip `CUSTOM` indices because we cannot recreate them without the custom Java class with the implementation. The name of the class could potentially contain sensitive information too. We can probably go a step further in a future PR to only remove custom indexes that do not contain a C* indexing class.

python/adelphi/adelphi/cql.py
```diff
@@ -1,7 +1,7 @@
 # Logic necessary to generate a string representation of a standard CQL schema
 
 from adelphi.anonymize import anonymize_keyspace
-from adelphi.store import get_standard_columns_from_table_metadata, set_replication_factor
```
grighetto, line 4: not used

python/adelphi/integration-tests/adelphi-docker-tests.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
```
grighetto, line 1: This is the main shell script for the integration tests, it starts up the Docker containers for each C*, populates the base schema, runs the anonymizer and compares the output with the expected result.

This is mostly here to guarantee that when we upgrade the Python driver, things will still work as expected and we won't leak any newly-added metadata without noticing.

python/adelphi/integration-tests/adelphi-docker-tests.sh
```diff
@@ -0,0 +1,78 @@
...
+function anonymize() {
+	local version=$1
+	echo "Anonymizing..."
+	adelphi export-cql > output/$version.cql 2>> logs/$version.log
+}
+
+function compare() {
+	local version=$1
+	echo "Comparing..."
+	diff schemas/$version.cql output/$version.cql
```
grighetto, line 43: relies on `diff` command (I think most Linux distros have it, but we can always install it in the CI server, when we get to that)

python/adelphi/integration-tests/adelphi-docker-tests.sh
```diff
@@ -0,0 +1,78 @@
...
+	diffExitCode=$?
```
grighetto, line 44: keeps track of the exit code, such that even if the last comparison succeeded, but some previous comparison failed, we want the parent script to fail

python/adelphi/integration-tests/base-schema.cql
```diff
@@ -0,0 +1,211 @@
+CREATE KEYSPACE my_ks_0 WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '1'}  AND durable_writes = true;
```
grighetto, line 1: this is the base schema that gets executed in all C* versions

it's generated with: `python3 schema_util.py > ../../integration-tests/base-schema.cql`

python/adelphi/tests/unit/schema_util.py
```diff
@@ -0,0 +1,106 @@
+from cassandra.metadata import Metadata,\
+	KeyspaceMetadata,\
+	TableMetadata,\
+	ColumnMetadata,\
+	IndexMetadata,\
+	SimpleStrategy, \
+	UserType
+
```
grighetto, line 8: utility script that programmatically generates a C* schema compatible with C* from 2.1.22 to 4.0-beta3

it contains a couple of keyspaces, tables, indexes (regular and custom) and complex UDTs

python/adelphi/tests/unit/test_anonymize.py
```diff
@@ -0,0 +1,133 @@
...
+from adelphi import anonymize
+from adelphi.anonymize import COLUMN_PREFIX, get_name
+from schema_util import get_schema
+
+try:
+    import unittest2 as unittest
+except ImportError:
+    import unittest  # noqa
+
+class TestCqlAnonymize(unittest.TestCase):
```
grighetto, line 11: anonymizer unit tests

grighetto: we can expand this to test the command-line params too, for now these gravitate towards the anonymization

python/adelphi/adelphi/anonymize.py
```diff
@@ -1,12 +1,10 @@
+import re
```
absurdfarce, line 1: Nit: this should be moved below the top-level comment for this package.  At least that's the convention used in the other Adelphi packages.

python/adelphi/adelphi/store.py
```diff
@@ -54,7 +54,6 @@ def partition(pred, iterable):
         log.info("Excluding system keyspaces " + ",".join((ks.name for ks in failed)))
     return passed
 
-
```
absurdfarce, line 57: Nit: there are two spaces between the various functions in all the Python packages.  We can change that if desired but a one-off change doesn't seem like a good idea.

python/adelphi/adelphi/anonymize.py
```diff
@@ -38,59 +33,70 @@ def get_name(original_name, prefix):
...
+
     for table in keyspace.tables.values():
         anonymize_table(table)
 
-    for udt in keyspace.user_types.values():
-        anonymize_udt(udt)
+    # remove functions, aggregates and views for now
+    keyspace.functions = {}
+    keyspace.aggregates = {}
+    keyspace.views = {}
```
absurdfarce, line 47: Is the removal of all of these required for anonymization?  Seems like we're giving up a fair amount of useful schema info here.

python/adelphi/adelphi/anonymize.py
absurdfarce: There's an explicit assumption here that name_map[TYPE_PREFIX] exists at this point, something that get_name does _not_ assume.  That assumption holds, but it only holds because of something that happens outside of this function... indeed, something that happens at a level _above_ this function.  That happens when we loop over all UDTs and populate their names first before this function is called.  All of that happens in anonymize_udts.

This sort of unenforced and undocumented dependency is almost always a recipe for silent breakage on future changes.  It would be nice if at a minimum we could make that dependency more explicit... and ideally remove it all together.

python/adelphi/adelphi/anonymize.py
```diff
@@ -38,59 +33,70 @@ def get_name(original_name, prefix):
...
-        clustering_key.name = name_map[CLUSTERING_KEY_PREFIX][clustering_key.name]
+    # remove comment since it may contain sensitive information
+    table.options["comment"] = ""
 
+    # The CQL python driver holds the same object references for clustering keys
+    # and regular columns, so we have to separate them otherwise the names
+    # will be re-anonymized.
     for column in get_standard_columns_from_table_metadata(table):
-        anonymize_column(column, COLUMN_PREFIX)
+        anonymize_column(column)
```
absurdfarce, line 96: Since we've now eliminated the distinction between clustering and partition keys and regular columns in an anonymization context can't we just simplify this logic to something like:

```python
for column in table.columns:
    anonymize_column(column)
```

python/adelphi/adelphi/anonymize.py
absurdfarce: Yeah, I was gonna mention the same thing... _some_ indexes we could re-create.  But I'm not sure it's a huge problem... waiting until a future release doesn't seem unreasonable.  We'll lose _some_ information about the use of these "custom" indices for the time being but that seems like a lower priority concern than getting the schemas in the first place.

absurdfarce: Perhaps create a ticket for this future work now, though?

python/adelphi/integration-tests/adelphi-docker-tests.sh
absurdfarce: More thinking out loud here than anything... not sure if this is a fully-formed thought or not yet.

I'm wondering if this shouldn't be converted to a Python script, more specifically a Python test, to facilitate running on Jenkins.  At the moment we get a binary response from this script in the form of an exit code: either everything passed or something failed.  Integrating with the Python testing frameworks would allow us to provide responses for each distinct C* version.

Not sure it's essential... like I said above, more thinking out loud here.

python/adelphi/integration-tests/adelphi-docker-tests.sh
absurdfarce: I would reasonably expect diff to be available in any context where these tests were run

python/adelphi/adelphi/anonymize.py
grighetto: @absurdfarce do you think we can safely anonymize functions and aggregates?

Maybe materialized views could be added, but since views aren't recommended for prod, it seemed like a lower priority that shouldn't block the rest of the PR. We can open a ticket to track this though.

python/adelphi/adelphi/anonymize.py
grighetto: > can't we just simplify this logic to something like

not really, because the python driver references the same objects in `table.columns` and `table.clustering_key`, so some columns would go through the anonymization twice. I added a comment in the code trying to clarify that.

Maybe we could skip the anonymization for `table.clustering_key` all together and simply rely on the `table.columns`, but if something changes internally in a future version of the driver and they don't have the same references anymore, we'll fail to anonymize some names. It seemed this was the most future-proof alternative.

python/adelphi/integration-tests/adelphi-docker-tests.sh
grighetto: > Integrating with the Python testing frameworks would allow us to provide responses for each distinct C* version

That's a good point! I'll look into that, but maybe it could be a follow-up PR.

python/adelphi/adelphi/anonymize.py
absurdfarce: My inclination is to say we probably can do this, but yeah, it can be held off for future work.  It does seem like including as many of these as possible in some capacity is probably worth it, though, since it will make for more robust validation testing.

python/adelphi/adelphi/anonymize.py
absurdfarce: Yeah, that second option is what I was after; don't do anything for clustering keys and just iterate through columns.  I agree that future changes to the driver could possibly break that but the only assumption we're making there is that table.columns gives us an iterator for every column in the table (regardless of it's role).  Breaking _that_ contract would be a significant change for a large number of users of the Python driver... so it seems like we're on pretty safe ground there.

Plus, I mean... it's not like there are a slate of impending Python driver changes going forward. :)

_And_ we're looking at pinning the driver version with the compatible operator, and since any future driver version that _did_ make the change mentioned above would _at least_ need a new minor version (more correctly a new major version for something so drastic) we'd be well insulated from future change.

So I guess I'd argue we're on pretty safe ground here. :)

python/adelphi/integration-tests/adelphi-docker-tests.sh
absurdfarce: Oh, absolutely.... this doesn't need to be done right now.  I'm :100: okay with handling this in a future ticket.

python/adelphi/adelphi/anonymize.py
absurdfarce: FTR I've created https://github.com/datastax/adelphi/issues/61 to track this future work

python/adelphi/integration-tests/adelphi-docker-tests.sh
absurdfarce: I've created https://github.com/datastax/adelphi/issues/62 to track this future work

absurdfarce:
pytest is actually a third-party tool for running Python unit tests (see https://docs.pytest.org/en/stable/).  It needs to be installed as a discrete package for use here, so make sure you do "pip install pytest" if you wish to use it.

python/adelphi/tests/unit/test_anonymize.py
```diff
@@ -0,0 +1,133 @@
+import pytest
```
absurdfarce, line 1: This import doesn't appear to be used: everything below uses the conventional Python unittest framework (which seems like the right decision to me) and I don't see any difference in behaviour when this import is removed.

python/adelphi/tests/unit/test_anonymize.py
```diff
@@ -0,0 +1,133 @@
...
+		self.assertTrue(table.options["comment"] == "Lorem ipsum")
+		anonymize.anonymize_table(table)
+		self.assertTrue(table.options["comment"] == "")
+
+	def test_anonymous_name_is_consistent(self):
+		name1 = get_name("test_column", COLUMN_PREFIX)
+		name2 = get_name("test_column", COLUMN_PREFIX)
+		name3 = get_name("another_column", COLUMN_PREFIX)
+		self.assertEqual(name1, name2)
+		self.assertNotEqual(name1, name3)
```
absurdfarce, line 133: Worth adding this at the end so that we can run this thing with a simple `python test_anonymize.py` (i.e. to avoid any explicit dependence on pytest):

```python
if __name__ == "__main__":
    unittest.main()
```

absurdfarce:
The unit test is currently failing when run against Python 2.7.14.  Looks mainly like a collection of mismatches due to the integers in  use for anonymized names.  It's not immediately clear to me if that process just isn't deterministic or if there's something else going on here.

python/adelphi/adelphi/anonymize.py
grighetto: > And we're looking at pinning the driver version with the compatible operator

your PR that got merged recently didn't pin to the exact version, right? it gives some flexibility to upgrade to a newer patch version right?

Anyway, I'm leaning towards leaving it the way it is since it reduces the risk of leakage to zero and removing it doesn't buy us much.

absurdfarce: Yeah, the PR uses the "compatible" operator which amounts to allowing different patch versions with the same major/minor.  It's true that this does allow the version to change, but you'll only have changes in the patch version... and by definition a patch change won't introduce API changes.  So I guess I see the risk here as _extremely_ low.

Broadly speaking I agree, it isn't a _huge_ thing... I'm mostly interested in trying to simplify the code as much as possible.  Looking at this code six months from now one could very credibly wonder why we're still looping over these various sets of columns instead of... just looping over the columns.  It's not gonna make or break anything, it's just simpler.

grighetto: > and by definition a patch change won't introduce API changes

Since we're mutating internal state of the Metadata object (which in general is supposed to be read-only), I'm not really sure if that qualifies as an API. I think that's mostly where my concern is.

absurdfarce: Yeah, I'm... not really thrilled with that, to put it mildly.  FWIW the second Github PR sought to address this by changing adelphi.store.with_cluster to call the passed cluster_fn with a cluster connection, close that connection and return the object.  So even though we're dealing with a Metadata instance we're not retaining a conn to the cluster anymore... so there's no obvious way for that data to be persisted.

We _are_ still modifying the Metadata object itself locally, which I agree isn't great... I really wish that object were immutable.  But that's a different concern from what we were discussing originally (i.e. column traversal).

We can leave it as-is and maybe consider taking a look at it later.  As I think we've both mentioned above it isn't by any means essential that we change this now; we can always revisit.

grighetto: Right, we grab the Metadata and close the connection, that's how it works from the beginning.

In my comment I wasn't concerned about the immutability nor the data being persistent, what I was trying to point at was the fact that by changing the internal state directly it doesn't respect any API contract, in other words, there's no guarantee the internal object references won't change even in a patch release, which would potentially break the column traversal.

Anyway, I still think the leakage of data is too big of a problem to run any risk around it, I prefer to have the anonymization completely under our control.

python/adelphi/adelphi/anonymize.py
grighetto: Notice this is not only used for UDTs, it's also used by `anonymize_column`, so this is here mostly to avoid code duplication.
I think the only way to make the dependency more explicit would be to lump everything together in `anonymize_keyspace` and accept some code duplication.

> That happens when we loop over all UDTs and populate their names first before this function is called.
> This sort of unenforced and undocumented dependency is almost always a recipe for silent breakage on future changes. 

If someone ever changes that code path to anonymize columns or fields before UDTs, the unit tests included in this PR will immediately fail.

python/adelphi/adelphi/anonymize.py
grighetto: fixed

python/adelphi/adelphi/store.py
grighetto: fixed

python/adelphi/adelphi/anonymize.py
grighetto: > My inclination is to say we probably can do this

Did you consider anonymizing the function body across all the supported languages? And even so, I think hiding the implementation of an UDF algorithm is more important than just anonymizing its variable names.
Anyway, that's a much larger scope and we can revisit it the future as you said. +1

python/adelphi/tests/unit/test_anonymize.py
grighetto: fixed

python/adelphi/tests/unit/test_anonymize.py
grighetto: done

grighetto:
@absurdfarce Thanks for the thorough review. PR feedback has been addressed.
I made the change to use `pip install` instead of `python setup.py install` to cope with your virtualenv setup (and probably other people would run into that too). Let me know how that goes for you now so we can move on with this PR.

python/adelphi/adelphi/anonymize.py
absurdfarce: Understood... I'd just prefer for the code to be more explicit about what it's doing there and how it handles those cases.  But given time constraints I don't feel like this is essential enough to hold up getting the larger set of improvements provided by this PR in so let's just skip it for now.

python/adelphi/adelphi/anonymize.py
absurdfarce: I still disagree, but not strongly enough to hold up the PR further... and it's something of a speculative concern anyway.  We can revisit in the future if needed.

FWIW my initial comment above (re: going back to get connection then returning Metadata) referred to an explicit change that happened in the first Github PR.  At some point prior to that (I forget exactly when) I became increasingly uncomfortable with the returned Metadata instance and changed the relevant logic.  Everything (evaluating the Metadata object, building up state, and printing the output) now happened within an input cluster_fn which took the Cluster instance as an arg; there was no return statement out of that logic.

Then I got nervous about setting RF on a Metadata object while still connected to the cluster and changed it back.  That's what you see at https://github.com/datastax/adelphi/pull/51/files#diff-03a41fd89b9e4d86cb31cefbd02da1b19aed39fb8dc9cbaea15f80b672c25d42

I mention this only to say that it wasn't _always_ the case that we return a Metadata object from the equivalent of store.with_cluster.  _That's_ why I went into the detail in my original comment.

absurdfarce:
@grighetto I agree that all the comments on this PR are put to bed so we're very close.  At this point the only thing holding this PR up from going in is the test failures: I continue to see unit test failures for both Python 2.7.x (2.7.14 specifically) and 3.x (3.9.0 specifically).  I'd like to at least see passing tests before we move this forward.  Is this something you can take a look at?

absurdfarce:
I'll report the error output I'm seeing from the unit tests just so there's a record of those errors somewhere.

After doing a "pip uninstall; pip install" of this package I'm observing the following failures using Python 2.x:

[python2_unit_test_failures.txt](https://github.com/datastax/adelphi/files/5662086/python2_unit_test_failures.txt)

A similar process (pip uninstall; pip install; manual run of unit tests) looks to pass without errors when using Python 3.x (specifically Python 3.9.0).

grighetto:
Tests fixed for Python 2 - it was caused by different dict sorting behavior across Python versions. Manually sorting the keyspace elements by name makes it deterministic and fixes the issues across the board.

absurdfarce:
Recent changes to get to passing unit tests seem sensible to me... I think we're finally done with this guy!

absurdfarce (approved):

Issue #36, opened 2020-10-22
Provide sample values.yaml files that are env specific


grighetto:
Some settings change depending on the environment (for example, the default storageClass for k3d is `local-path` whereas for GKE it is `standard`).
We can provide some sample config files that are tuned for the environment the user is targeting (local or cloud).

┆Issue is synchronized with this [Jira Task](https://datastax-oss.atlassian.net/browse/AD-87) by [Unito](https://www.unito.io)
┆Issue Number: AD-87

PR #56, opened 2020-11-24
#32 Build and publish Docker images for dependencies

grighetto:


Docker/build-images.sh
```diff
@@ -0,0 +1,19 @@
...
+}
+
+# try to find Spark's Dockerfile, if it doesn't exist, download the Spark tarball
+if [ ! -f "$SPARK_DOCKERFILE" ]; then
+	download_spark
+fi
+
+docker build --tag spark:3.0.0 -f $SPARK_DOCKERFILE spark
+docker build --tag cassandra-diff:4c9bc4 cassandra-diff
+docker build --tag gemini:1.7.3 gemini
```
grighetto, line 19: Bash script to build the images from the newly added Dockerfiles.
Spark comes with a Dockerfile, so we don't have to build a custom one.

The cassandra-diff pulls from the Spark image and that's what we publish to Dockerhub. There's no need to publish Spark by itself.

Docker/push-images.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# Dockerhub org
+REPO="adelphidev"
```
grighetto, line 4: We have to replace this by the final dockerhub account name we'll use (when we know what that is).

helm/adelphi/templates/gemini-job.yaml
```diff
@@ -42,12 +42,14 @@ spec:
...
-      workingDir: /workspace
+      # remove this container when Adelphi is available in PyPI
+    - name: clone-adelphi-cli
+      image: alpine/git
+      command:
+        - sh
+        - -c
       args:
-        - wget -O - {{ .Values.gemini_download_url }} | tar -zxvf -
+        - git clone https://github.com/datastax/adelphi.git /workspace/adelphi
```
grighetto, line 52: temporary solution while Adelphi is not in PyPI

helm/adelphi/values.yaml
```diff
@@ -15,10 +15,6 @@ target:
   version: "4.0.0"
   commit_hash:
 
-cassandra_diff_download_url: https://repository.apache.org/service/local/repositories/snapshots/content/org/apache/cassandra/diff/spark-uberjar/0.2-SNAPSHOT/spark-uberjar-0.2-20200325.160556-1.jar
-spark_download_url: https://archive.apache.org/dist/spark/spark-3.0.0/spark-3.0.0-bin-hadoop3.2.tgz
-gemini_download_url: https://github.com/scylladb/gemini/releases/download/v1.7.3/gemini_1.7.3_Linux_x86_64.tar.gz
```
grighetto, line 20: not needed anymore, we'll simply use the images

helm/adelphi/values.yaml
```diff
@@ -1,8 +1,8 @@
 namespace: cass-operator
-clusterSize: 3 # number of nodes per cluster
+clusterSize: 1 # number of nodes per cluster
 allowMultipleNodesPerWorker: false
 
-storageClassName: standard
+storageClassName: local-path
```
grighetto, line 5: back to k3d by the default

grighetto:
@absurdfarce I removed the NFS server because it really doesn't play well with kubelet's internal DNS and the only workaround would be replacing the NFS service name by a static IP, which is not good. The NFS service worked on GKE because it relied on the cloud DNS instead.

We'll have to come up with a proper storage option for the results data, I'm thinking we'll have to make it configurable in the Helm chart and we can propose some sane defaults through #36.

helm/adelphi/templates/gemini-job.yaml
```diff
@@ -57,8 +59,9 @@ spec:
       args:
         - echo "Extract Gemini schema using anonymizer script";
           pip install cassandra-driver;
-          wget -q https://raw.githubusercontent.com/datastax/adelphi/master/schema-anonymizer/anonymizer.py;
-          python anonymizer.py --hosts="{{ .Values.source.clusterName }}-{{ .Values.source.dc }}-service" --username="{{ .Values.source.clusterName }}-superuser" --password="`cat /source-secret/password`" --keyspaces="{{ .Values.gemini_schema_keyspace }}" --format=gemini --no-anonymize > /workspace/gemini_schema.json;
+          cd /workspace/adelphi/python/adelphi;
+          python3 setup.py install;
```
absurdfarce, line 63: Semi-nit: probably prefer "pip install ." here (along the same lines as similar discussions on other PRs).

Docker/gemini/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM golang
+
+ENV GEMINI_URL="https://github.com/scylladb/gemini/releases/download/v1.7.3/gemini_1.7.3_Linux_x86_64.tar.gz"
+
+RUN wget -O - $GEMINI_URL | tar -zxvf -
+
+CMD /gemini
```
absurdfarce, line 7: Am I nuts or does this not match up to what's in gemini-job.yaml?  It looks like in that case the binary is installed as /go/gemini but here it appears to be at /gemini... ?

helm/adelphi/templates/gemini-job.yaml
```diff
@@ -9,14 +9,14 @@ spec:
...
       command:
         - sh
         - -c
       args:
         - mkdir /results/gemini;
-          /workspace/gemini
+          /go/gemini
           --oracle-cluster={{ .Values.source.clusterName }}-{{ .Values.source.dc }}-service
           --oracle-username={{ .Values.source.clusterName }}-superuser
           --oracle-password="`cat /source-secret/password`"
```
absurdfarce, line 22: Is there a clean way to move all of these args into the Dockerfile as well?  Might be nice to avoid having to spell all this out within the YAML if we can do so without creating a configuration nightmare.

Docker/build-images.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
absurdfarce, line 1: Semi-nit: probably prefer "docker" to "Docker" for this directory name.  Everything else underneath the root "adelphi" directory is lower-case and I would think we'd want to preserve that here.

absurdfarce:
@grighetto This looks pretty much like what I was expecting: nice work!  This will be a definite improvement once we get this in.

I'm still looking at a few things, but I was wondering in general... do you have a simple recipe for testing the workflow using the newly-generated images?  I suppose something like this would work:

* Deploy a local registry
* Use the included script to build the images
* Use the included script to deploy the images to the local registry
* Run the workflow pointing it at the local registry

Do you have any additional guidance and/or a Better Way :tm: to do something like that?

absurdfarce:
Oh, and before I forget (again)... point taken on the NFS removal.  I actually kinda like that change because it feels like we've simplified things quite a bit... while obviously also giving up some important functionality.  But those NFS configs definitely brought some overhead with them.

As for a general solution... I kinda like the idea of proposing some ideas via the sample charts.  It also occurred to me earlier that we don't necessarily need a full shared filesystem to make that happen; we really only need the ability for the images that generate output to copy that output over to the results server.  We might be able to do something like that with something as simple as rsync.  This assumes a couple of things which may or may not be a problem:

* Nodes have some capacity to discover the results server
* Assuming we're using rsync over SSH we have some way to make the SSH creds work

I don't have enough k8s knowledge yet to know how big a deal those two things are... but assuming they aren't fatal it's something to consider.

Docker/gemini/Dockerfile
grighetto: No, you're not nuts, good catch actually, the right path should be `/go/gemini` which is the default workdir for the golang image. Originally I had a differente base image.

Fixed!

helm/adelphi/templates/gemini-job.yaml
grighetto: Here we're passing args directly to the gemini cli. We can't really hardcode them in the image.

Docker/build-images.sh
grighetto: Done

grighetto:
@absurdfarce You should be able to test this locally with something like this https://docs.docker.com/registry/deploying/
But you'll have to change the `push-images` script to append the local registry URL.
And you'll also have to change gemini-job.yaml and spark-job.yaml to point to the local registry.

grighetto:
As for the NFS server, it will probably make a come back. It doesn't seem there's a great solution out there yet to mount an object store as a filesystem that we can mount as a ReadWriteMany volume in k8s.
So, I think the final solution will be pretty similar to what we had before (with an workaround for the DNS issue) + the ability to override the volume config (in order to use something like Google Filestore for persistent storage). More on that in a future PR for #64

helm/adelphi/templates/gemini-job.yaml
grighetto: Fixed

docker/cassandra-diff/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+ENV GIT_ID="4c9bc4f4e3fd7d23b1284c89266ffbf10b8f0183"
+
+RUN git clone https://github.com/apache/cassandra-diff.git cassandra-diff; \
+	cd cassandra-diff; \
+	git checkout $GIT_ID
+
+WORKDIR /cassandra-diff
+RUN mvn clean package -DskipTests
+
+FROM adelphi-dev/spark:3.0.0
```
absurdfarce, line 14: I'm pretty sure this needs to be:

```
FROM spark:3.0.0
```

instead.  I'm assuming the intent here was to reference the version of the Spark image that's built by build-images.sh just before we try to build this guy, but that's just saved as spark:3.0.0.

helm/adelphi/templates/gemini-job.yaml
absurdfarce: Now that we're there presumably all of this can come out... ?

absurdfarce:
A small reminder (probably just to me) for testing purposes: make sure to delete any pre-built instances of these images you may already have locally before running build-images.sh:

```
docker rmi spark:3.0.0
docker rmi gemini:1.7.3
docker rmi cassandra-diff:4c9bc4
```

It's important to make sure we have everything we need to build this thing entirely from scratch

helm/adelphi/templates/results-server.yaml
```diff
@@ -1,33 +1,34 @@
...
\ No newline at end of file
+    initContainers:
+    - name: clear-index
+      image: alpine
+      command:
+        - sh
+        - -c
+      args:
+        # remove sample index created by the NFS server
+        - rm /results/index.html
```
absurdfarce, line 27: I was seeing errors with this due to /results/index.html, perhaps just on an initial run before anything had been generated (I'm assuming the persistent volume would address that for future runs).  I think just switching this to "rm -f /results/index.html" would be adequate to address it.

docker/cassandra-diff/Dockerfile
absurdfarce: Yup, sure looks like it.  If I try to build as this sits on a newly-initialized OpenStack node:

```
Step 6/7 : FROM adelphi-dev/spark:3.0.0
pull access denied for adelphi-dev/spark, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

Just changing to "spark" directly resolves the issue.

jdonenine: Is this one we actually want to come from the adelphitools org we created?

I don't think we've published a spark image though.

absurdfarce: Nah, as you mentioned @jdoneine there isn't a Spark image up there.  We build the Spark image locally in build-images.sh so we actually want the build of cassandra-diff to use _that_ image... which is why I argued for "spark:3.0.0" (i.e. "use my local build of the Spark 3 image").

helm/adelphi/templates/results-server.yaml
absurdfarce: Will address this in small follow-up PR

docker/cassandra-diff/Dockerfile
absurdfarce: Will address this in small follow-up PR

helm/adelphi/templates/gemini-job.yaml
absurdfarce: Will address this in small follow-up PR

docker/cassandra-diff/Dockerfile
jdonenine: Ah, gotcha.  Sorry, this is the building workflow itself.  Ignore me and move along 😄

absurdfarce:
Testing has shown that the changes here represent a significant improvement in the workflow both in terms of stability and performance.  We expected this and it's good to see it hold true in practice.  Well done @grighetto !

There are a few nits in the comments above but none of them are anywhere near significant enough to hold off from moving forward with this... I also selfishly want to start using the image-enabled workflow as the default.  To expedite that process I'm going to merge this PR as it stands and address the few quibbles in a smaller follow-up PR which @grighetto can review when he's back.

absurdfarce (approved):

absurdfarce:
Follow-up PR lives at https://github.com/datastax/adelphi/pull/84

Issue #40, opened 2020-10-28
Allow for uploading schemas straight from the anonymizer into the repository


grighetto:
While discussing a few ideas with @absurdfarce, he suggested that we allow uploading schemas straight from the anonymizer to make the process even more straightforward.

┆Issue is synchronized with this [Jira Task](https://datastax-oss.atlassian.net/browse/AD-91) by [Unito](https://www.unito.io)
┆Issue Number: AD-91

josh-mckenzie:
So is the thinking is we have an external repository that users have write-only access w/a submission API w/a token or some kind of UID they can use to then later upsert it? Probably need to lean on token after initial upload required to also get the "auth, don't overwrite" side of it covered... Hm.

Way better UX than "submit a PR" though, I must say.

absurdfarce:
The Github auth stuff has always seemed opaque to me, but as best I can tell the personal access token is bound to a specific user.  We only want to grant access to public repos for a single repo (our schema upload repo) but it looks like use of a personal access token grants access to _every_ repo (this should probably be confirmed).  So we'll want to constrain this in some way.

I _think_ we can create a new user which only has access to the schema upload repo and create the personal access token for that user... that should adequately constrain things.  We _may_ even be able to make the personal access token more public in that case... but I want to do some more testing to confirm that nefarious actors can't do bad things in that case.

I've also confirmed that the personal access token scheme + https://github.com/PyGithub/PyGithub at least supports feature branch creation + content upload.  It looks like PR creation follows on nicely from there... again, more testing required to say for sure.

PR #63, opened 2020-11-26
33 a little metadata

absurdfarce:


absurdfarce:
Apologies, this one got a little confusing, mainly because I am an idiot.

This started life as a PR off of the work being done for https://github.com/datastax/adelphi/issues/40 (specifically https://github.com/datastax/adelphi/pull/51).  When that issue was closed I made the mistake of deleting the feature branch _before_ rebasing the other PR off of master; that in turn effectively killed the first PR.  The simplest way to get back to a good state seemed to be to pull in the branch for the old PR itself ("git pull origin pull/55/head") and handle any last minute conflicts that may have come up.  The result of that process is what you see in this PR.

The "extra" commits referenced above presumably come from the fact that HEAD of pull/55 would have included the Github work (since I originally branched off of that).  Since those commits are now in master they should be silently ignored here.

The contents of this PR should match up to what's in the prior version (https://github.com/datastax/adelphi/pull/55) with a few changes based on last-minute changes in the issue 40 PR.

python/adelphi/bin/adelphi
```diff
@@ -29,25 +34,29 @@ import click
...
 @click.option('--password', help='Database password')
 @click.option('--keyspaces',
               help='Comma-separated list of keyspaces to include. If not specified all non-system keypaces will be included')
 @click.option('--rf', type=int, help='Replication factor to override original setting. Optional.')
-@click.option('--anonymize/--no-anonymize', default=True, help="Enable/disable schema anonymization")
-@click.option('--output-dir', help='Schema files will be written to this directory. If not specified schemas will be written to stdout.')
+@click.option('--anonymize/--no-anonymize', default=True, show_default=True, help="Enable/disable schema anonymization")
+@click.option('--output-dir', help='Directory schema files should be written to. If not specified, it will write to stdout')
+@click.option('--purpose', help='Comments on the anticipated purpose of this schema.  Optional')
+@click.option('--maturity', help='Comments on the maturity of this schema.  Optional')
```
grighetto, line 57: I'm wondering if we should offer some guidance here, like list some options in help: alpha, beta, prod, staging, ... I think it gives us a better chance of people actually filing this in and having a consistent format.
I'm not saying we should enforce any of these values, but simply propose them.
@absurdfarce wdyt?

grighetto (approved):
Pending question around maturity command help.
Otherwise +1

python/adelphi/bin/adelphi
absurdfarce: Per external conversation we're going to be doing exactly this as a middle ground between a free text field vs. an enum of allowed values.  Change to follow shortly.

Issue #67, opened 2020-12-02
Naming collision on schema files for generated PRs


absurdfarce:
Current naming convention for schema files is <username>/<keyspace name>.  After this was implemented we moved schema anonymization higher up in the process which means <keyspace name> is always "ks_0" (or something similar) at the time of file creation.  As a result we're very subject to collisions; the following should make it clear why:

* User contributes schema for keyspace foo, generates name someuser/ks_0
* This PR is merged
* User contributes schema for keyspace bar, generates name someuser/ks_0... but this time it's for an entirely different keyspace
* New PR cannot be merged since it conflicts with what's already in the repo

absurdfarce:
Simplest answer here is probably to record a SHA hash of the origin keyspace name and include it as an identifier for the keyspace.  This identifier can be used in place of the keyspace name (i.e. someuser/ksname-digest).  It should probably also be added to the keyspace file as metadata.

We should also consider adding the contributing Github user as metadata as well.

And perhaps a SHA of the actual contents of the keyspace dump as well.

absurdfarce:
In addition to fixing the collision issue mentioned above this PR adds the following bits of metadata to generated schemas:

* hashed_ks_name = what it says on the tin; a base64-encoded repr of the SHA256 hash of the keyspace name.  This provides a discrete "name" to use even for anonymized keyspaces
* created_at = ISO 8601 timestamp indicating when the schema was exported
* github-user = if contributing to Adelphi the Github user used to make that contribution is preserved

Issue #70, opened 2020-12-08
Repository.create_fork fails on Python 2.x if user is in a Github org


absurdfarce:
Documenting this here mainly so that there's a record of it.  Fix will likely be implemented as part of the ongoing work for https://github.com/datastax/adelphi/issues/67.

Observed while attempting to run "adelphi contribute" using Python 2.x after switching code to use the repo at datastax/adelphi-schemas (which is in the "datastax" organization):

```python
Traceback (most recent call last):
  File "/work/git/adelphi/python/setup/bin/adelphi", line 202, in <module>
    export(obj={}, auto_envvar_prefix="ADELPHI")
  File "/work/git/adelphi/python/setup/lib/python2.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/work/git/adelphi/python/setup/lib/python2.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/work/git/adelphi/python/setup/lib/python2.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/work/git/adelphi/python/setup/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/work/git/adelphi/python/setup/lib/python2.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/work/git/adelphi/python/setup/lib/python2.7/site-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/work/git/adelphi/python/setup/bin/adelphi", line 196, in contribute
    fork_repo = origin_repo.create_fork()
  File "/work/git/adelphi/python/setup/lib/python2.7/site-packages/github/Repository.py", line 2088, in create_fork
    "POST", self.url + "/forks", input=post_parameters,
  File "/work/git/adelphi/python/setup/lib/python2.7/site-packages/github/Requester.py", line 322, in requestJsonAndCheck
    verb, url, parameters, headers, input, self.__customConnection(url)
  File "/work/git/adelphi/python/setup/lib/python2.7/site-packages/github/Requester.py", line 413, in requestJson
    return self.__requestEncode(cnx, verb, url, parameters, headers, input, encode)
[...]

Issue #65, opened 2020-12-02
Consider adding a feature to "adelphi" app to allow a user to determine whether a schema has already been submitted


absurdfarce:
Recent changes have moved the anonymization process to an earlier point in the workflow of the "adelphi" app.  As a result the actual files in the generated Github PRs are named after anonymized keyspaces... which leaves a user no obvious way to determine if a keyspace they want to submit to Adelphi has already been submitted.  This ticket would address this deficiency.

We likely wouldn't need much more than (1) adding a SHA hash of the (anonymized) schema as metadata and (2) allowing some kind of ability to query metadata from the "adelphi" app.

One might also ask why bother using the anonymized keyspace name for the file name in the repo if it doesn't provide any useful information.  This question is entirely legit and is being explored elsewhere... it'll probably wind up in a ticket of it's own.  That work should be considered to be separate from this effort, however.  The focus of this ticket is very much on the ability to determine whether a schema has already been submitted.  They're certainly related, however; I can imagine the naming convention in the repo being changed to use the SHA hash of the schema and this being the basis for the query mechanism.

PR #71, opened 2020-12-08
#67 Fix for naming collisions + Python2 support for create_fork + other minor fixes

absurdfarce:


python/adelphi/bin/adelphi
```diff
@@ -42,6 +45,32 @@ from adelphi.store import with_cluster as _with_cluster
...
+            post_parameters["organization"] = organization
+        headers, data = self._requester.requestJsonAndCheck(
+            "POST", self.url + "/forks", input=post_parameters,
+        )
+        return Repository.Repository(self._requester, headers, data, completed=True)
+
+    import types
+    repo.safe_create_fork = types.MethodType(safe_create_fork, repo)
+
+
```
absurdfarce, line 73: The heart of the fix for #70

python/adelphi/bin/adelphi
```diff
@@ -54,7 +83,7 @@ log = logging.getLogger('adelphi')
 @click.option('--anonymize/--no-anonymize', default=True, show_default=True, help="Enable/disable schema anonymization")
 @click.option('--output-dir', help='Directory schema files should be written to. If not specified, it will write to stdout')
 @click.option('--purpose', help='Comments on the anticipated purpose of this schema.  Optional.')
-@click.option('--maturity', help='The maturity of this schema.  Sample values would include ‘alpha’, ‘beta’, ‘dev’, ‘test’ or ‘prod’.  Optional.')
+@click.option('--maturity', help="The maturity of this schema.  Sample values would include 'alpha', 'beta', 'dev', 'test' or 'prod'.  Optional.")
```
absurdfarce, line 86: The backticks were causing a problem for Python 2.x; it wasn't happy with the non-ASCII-ness of these characters.

python/adelphi/bin/adelphi
```diff
@@ -76,7 +105,7 @@ def export(ctx, hosts, port, username, password, keyspaces, rf, anonymize, outpu
     ctx.obj['port'] = port
     ctx.obj['username'] = username
     ctx.obj['password'] = password
-    ctx.obj['keyspace_names'] = keyspaces.split(',') if keyspaces is not None else None
+    ctx.obj['keyspace-names'] = keyspaces.split(',') if keyspaces is not None else None
```
absurdfarce, line 108: Part of an attempt throughout this PR to uniformly choose between underscores and hyphens when separating names.

python/adelphi/adelphi/gh.py
```diff
@@ -22,11 +22,6 @@
 
 from github import Github
 
-import logging
-import os
-import os.path
-import uuid
-
```
absurdfarce, line 29: Minor cleanup; we had duplicate imports here for no obvious reason

absurdfarce:
This PR addresses both https://github.com/datastax/adelphi/issues/67 and https://github.com/datastax/adelphi/issues/70

python/adelphi/bin/adelphi
```diff
@@ -103,104 +132,124 @@ def unique(iterable, key=None):
...
     else:
-        for keyspace_obj in keyspace_objs:
-            with open(os.path.join(export_dir, keyspace_obj.name),'w') as output_file:
+        for (keyspace_obj, hashed_ks_name) in keyspace_map.items():
+            # Output filename should use the file-safe alphabet for b64 encoding
+            output_file_name = hash_to_filename(hashed_ks_name) if props["anonymize"] else keyspace_obj.name
+            with open(os.path.join(export_dir, output_file_name),'w') as output_file:
                 log.info("Writing schema for keyspace " + keyspace_obj.name + " to file " + output_file.name)
+                # Metadata string for b64 encoded keyspace name should use the conventional encoding
+                metadata['hashed_ks_name'] = hash_to_str(hashed_ks_name)
```
absurdfarce, line 196: Note that "hashed_ks_name" is added to the metadata in export_keyspaces, meaning it's added every time we do an export regardless of whether it's CQL, Gemini and/or going or not going to Github.  Rationale here is that we want to use the hashed keyspace name as a unique identifier (at least within the domain of "schemas submitted _by this user_").

Further note that "github-user" is added in the actual "contribute" logic itself.  That metadata field has no meaning outside the notion of a Github submission so seemed sane to constrain it there.

This notion of different tiers of metadata is implicit here but probably worthwhile to point out.

absurdfarce:
The more I look at this the less I love the format for metadata used here:

```
//Schema metadata:
//@host-count = 1
//@dc-count = 1
//@github_user = absurdfarce
//@created_at = 2020-12-08T04:02:32.953997
//@hashed_ks_name = c/iVnEJm2FhMUbXjntLdR5rJZVxB/wXJqgZFGlEcAM4=
```

The use of comments in CQL (and JSON for that matter) seems right to me, but I'd like for the metadata to be something that we can easily parse in code; this would encourage better automated tooling along the lines of what's considered in https://github.com/datastax/adelphi/issues/65.  I don't necessarily know what the right answer is here.. but it seems like something we should clarify before we call this good.

grighetto:
@absurdfarce It's not clear to me why we can't consolidate to a single hashing function, that is, `hash_to_filename`.
Right now, if I run with `--no-anonymize` OFF, the file name and the keyspace name in the the metadata will be slightly different (because of the URL encoding), and a third variation is when I run with `--no-anonymize` ON because it will only hash but not `base64encode`. 

Wouldn't it be possible to use `hash_to_filename` at all times?

absurdfarce:
@grighetto My rationale for splitting things up like this was that the "urlsafe" hashing used for filenames isn't the normal base64 people are used to.  If we assume we need a keyspace_id to uniquely identify a keyspace (currently the base64-encoded SHA256 hash of the keyspace name) _and_ we assume that we'll want to write a directory or file with that name then we can't use the stock base64 alphabet since it contains some chars used by filesystems.  Thus the "urlsafe" variant seems almost required.

This in turn made me concerned precisely _because_ it isn't the normal base64 alphabet.  I was concerned that users might look at it and believe it to be base64 while proving unable to match it after hashing their own directory names.  More importantly the process for generating urlsafe base64 encodings is Python-specific, so any tooling that gets created around generating and/or validating these names would either have to be in Python or re-implement the functionality.  Given all of that it seemed worthwhile to display the "pure" base64 encoded hash (using the standard base64 alphabet) in the metadata.

I can see an argument which says "base64 is an implementation detail, the user shouldn't concern themselves with that"... but I think we're kidding ourselves if we say that.

grighetto:
@absurdfarce having 3 different encodings for the same value doesn't seem very helpful to me.

> we can't use the stock base64 alphabet since it contains some chars used by filesystems. Thus the "urlsafe" variant seems almost required.

Right, that's exactly what I was advocating for, to use "urlsafe" all around.

> More importantly the process for generating urlsafe base64 encodings is Python-specific, so any tooling that gets created around generating and/or validating these names would either have to be in Python or re-implement the functionality.

That doesn't sound too bad (if it ever becomes a need), all it's doing is replacing `+` and `/` by `-` and `_`, respectively. https://github.com/python/cpython/blob/master/Lib/base64.py#L111

As things stand now, if an user runs Adelphi with `no-anonymize` locally to confirm if he has already contributed that schema, the `hashed_ks_name` won't match with what's in the git repo.

grighetto:
And since we have pytest on master now, I think this could use some unit testing too, but I'd be ok with a follow-up PR.

grighetto (approved):
Not quite onboard with the hashing approach, but I don't see a reason why we couldn't merge this otherwise.

python/adelphi/bin/adelphi
```diff
@@ -15,9 +15,13 @@
 # limitations under the License.
 
 import argparse
+import hashlib
 import logging
 import os
 import os.path
+import sys
+from base64 import urlsafe_b64encode
+from datetime import datetime, tzinfo, timedelta
```
absurdfarce, line 24: Imports here required for the Python2 UTC workaround below

python/adelphi/bin/adelphi
```diff
@@ -30,6 +34,20 @@ try:
...
+    class UTC(tzinfo):
+        def utcoffset(self, dt):
+            return timedelta(0)
+        def tzname(self, dt):
+            return "UTC"
+        def dst(self, dt):
+            return timedelta(0)
+    utc = UTC()
+
+
```
absurdfarce, line 50: Another Python2/3 incompatibility.  Python 2.x doesn't have datetime.timestamp.utc so we make a fake one if we can't find one provided by the platform.  Impl here based on https://stackoverflow.com/questions/7065164/how-to-make-an-unaware-datetime-timezone-aware-in-python

python/adelphi/bin/adelphi
```diff
@@ -103,104 +151,119 @@ def unique(iterable, key=None):
...
-        for keyspace_obj in keyspace_objs:
-            with open(os.path.join(export_dir, keyspace_obj.name),'w') as output_file:
+        for (keyspace_obj, keyspace_id) in keyspace_map.items():
+            # Output filename should use the file-safe alphabet for b64 encoding
+            output_file_name = keyspace_id if props["anonymize"] else keyspace_obj.name
+            with open(os.path.join(export_dir, output_file_name),'w') as output_file:
                 log.info("Writing schema for keyspace " + keyspace_obj.name + " to file " + output_file.name)
+                metadata['keyspace_id'] = keyspace_id
                 output_file.write(export_fn([keyspace_obj], metadata, options))
+                del metadata['keyspace_id']
```
absurdfarce, line 209: I prefer the name "keyspace_id" here and converted to it throughout this PR, not just in metadata but also in code.  It moves away from a name based on the impl (the various "hash" names that were used before) and identifies the field based on it's function in the code, which seems better on it's face.

python/adelphi/bin/adelphi
```diff
@@ -103,104 +151,119 @@ def unique(iterable, key=None):
...
+    return {"host-count": len(hosts), "dc-count": sum(1 for i in unique_dcs)}
 
-    validate_keyspaces(keyspaces)
-    return keyspaces
 
+def build_keyspace_map_and_metadata(props):
+    def build_fn(cluster):
+        metadata = {k : props[k] for k in ["purpose", "maturity"] if k in props}
+        metadata.update(get_cluster_metadata(cluster))
+        metadata["creation_timestamp"] = datetime.now(utc).isoformat()
```
absurdfarce, line 189: A timestamp for schema creation has general utility.  This was being created in the Github contribute logic but I promoted it here so that every schema will include it as metadata.

absurdfarce:
Good news, @grighetto : after prompting from this conversation I did a bit of digging and discovered that the "urlsafe" impl provided by Python actually _is_ formalized in a spec.  It's explicitly defined in [section 5 of RFC 4648](https://tools.ietf.org/html/rfc4648#section-5) which completely nullifies my concerns above.  If the custom base64 alphabet used by Python is more generic there isn't any concern about generating encoded names in a Python-specific format.

As such I've consolidated everything to the urlsafe encoding format.  This impl has been tested against both Python2 and Python3 and seems to work as expected.

If possible would you mind taking _one more pass_ to confirm that you're good with these changes?

grighetto (changes requested):
Integration tests are failing now:

```
2,3c2,4
< //@host_count = 1
< //@dc_count = 1
---
> //@host-count = 1
> //@dc-count = 1
> //@creation_timestamp = 2020-12-15T15:31:43.144602+00:00
```

It looks like there's some inconsistencies with the usage of `-` and `_`. I realize there was an one-off change to make this uniform in this PR.

I also noticed there's a behavior difference when exporting the schema to a file vs printing to console. When printing, it doesn't include the `keyspace_id`, so that's not covered by the integration tests.

Last thing, the keyspace_id in an anonymized schema is still different from the one in the non-anonymized schema, for the same keyspace.

absurdfarce:
Grrrr, also just noticed on that branch:

```
(python3-venv) [mersault@localhost python]$ adelphi --keyspaces=testkeyspace,system_views --output-dir=/tmp/foo --maturity=dev export-cql
Usage: adelphi [OPTIONS] COMMAND [ARGS]...
Try 'adelphi --help' for help.

Error: Invalid value for '--anonymize': anonymization enabled is not a valid boolean
```

Looks to be a side effect of the change in defaults in the other PR.  I'll see if I can decouple the default text from the actual default value.

absurdfarce:
A quick set of comments on the issues raised from the last review by @grighetto + the click issue I referenced earlier:

1. Fixed the issue with default value/help text around anonymize/no-anonymize.  Turns out the click lib allows us to just specify a flag, and a flag for "no-anonymize" is exactly what we want here (anonymize if not present, don't if it is).  The help text looks right too.

2. The issue with disconnects between "-" and "_" is indeed caused by renames in this PR; the canonical results for the integration tests haven't been updated.  That'll all change in the metadata format change PR anyway so I'm not gonna worry about it here.

3. keyspace_id was left out of the output when printing to standard out because we can request more than one keyspace here... and we haven't decided how to render that in the metadata scheme.  I'll prolly defer this to the metadata re-work ticket as well.

4. To be clear about the underlying problem re: keyspace_id when anonymizing vs. not anonymizing:

```
(python3-venv) [mersault@localhost python]$ adelphi --keyspaces=testkeyspace,system_views --output-dir=foo export-cql
...
INFO:adelphi:Writing schema for keyspace ks_0 to file foo/c_iVnEJm2FhMUbXjntLdR5rJZVxB_wXJqgZFGlEcAM4=
(python3-venv) [mersault@localhost python]$ adelphi --keyspaces=testkeyspace,system_views --output-dir=foo --no-anonymize export-cql
...
INFO:adelphi:Writing schema for keyspace testkeyspace to file foo/testkeyspace
(python3-venv) [mersault@localhost python]$ ls foo
'c_iVnEJm2FhMUbXjntLdR5rJZVxB_wXJqgZFGlEcAM4='   testkeyspace
(python3-venv) [mersault@localhost python]$ head -5 foo/c_iVnEJm2FhMUbXjntLdR5rJZVxB_wXJqgZFGlEcAM4\= 
//Schema metadata:
//@host-count = 1
//@dc-count = 1
//@creation_timestamp = 2020-12-16T07:38:04.426210+00:00
//@keyspace_id = c_iVnEJm2FhMUbXjntLdR5rJZVxB_wXJqgZFGlEcAM4=
(python3-venv) [mersault@localhost python]$ head -5 foo/testkeyspace 
//Schema metadata:
//@host-count = 1
//@dc-count = 1
//@creation_timestamp = 2020-12-16T07:38:20.566828+00:00
//@keyspace_id = 9oeoJFuUPUpGOIjokdIIuwsJcmGny03H8MyjkovpV4w=
```

The problem here is https://github.com/datastax/adelphi/blob/d93980b2e1f9d5df740eb439242bc97676359a60/python/adelphi/bin/adelphi#L176.  anonymize_keyspace(ks) has side effects; it modifies the input keyspace object rather than returning a modified copy.  As a result if the anonymize_keyspace() call happens before the build_keyspace_id() call when the dict comprehension is eval'd then we'd wind up feeding the _anonymized_ keyspace name into build_keyspace_id... which is exactly what we were hoping to avoid.

To confirm we can move build_keyspace_id() to an operation explicitly before the anonymize_keyspace() op and then re-test.  A quick test suggests that this does in fact fix our problem: both keyspaces now show a keyspace ID of "9oeo..." which this change.

grighetto:
@absurdfarce 
1. Looks good 👍 
2. `"the canonical results for the integration tests haven't been updated"` - I wasn't referring only to the integration tests, for example, we went from `keyspace_names` to `keyspace-names`, but also added `creation_timestamp`, so it widens the discrepancy. But no big deal, we can fix that later.
3. I think the suggestion I left #77 would handle this (that is, to process one keyspace at a time), but we handle later, it won't affect the contributions, so we're good.
4. Tested it and looks good now 👍 

I think the only outstanding issue is the broken Gemini printing. Do you intend to fix that in this PR?

absurdfarce:
Good point @grighetto !

I ported over the changes I referenced when we were discussing that issue elsewhere.  I tested these changes using export-cql + export-gemini (with and without anonymization) using Python2 + Python 3.9.0.  Seemed to work normally in all cases.

grighetto:
@absurdfarce that last commit broke Gemini output to file. I commented on that proposed solution here https://github.com/datastax/adelphi/pull/77#discussion_r544407997

absurdfarce:
Ah, sorry @grighetto , I think I misunderstood your comment on the other PR.  The picture is a little clearer to me now.

These changes do make things better by making explicit the idea that the export functions expect keyspaces to be supplied via an iterator.  That allows them to work to a single interface by putting conversion to that interface back onto the caller.

Now, all of _that_ said, this still isn't ideal; there's still too much customization in bin/adelphi that ideally would get moved out to the format-specific export fns.  Agreed that there's definitely future work there... I may even try to tackle that in the schema re-work.

grighetto:
> These changes do make things better by making explicit the idea that the export functions expect keyspaces to be supplied via an iterator. That allows them to work to a single interface by putting conversion to that interface back onto the caller.

That interface could take a single keyspace object too if we want, which would prevent us from creating an iterator with a single item in the caller.

grighetto (approved):
Tested latest changes and feature-wise all looks good now 👍

Issue #72, opened 2020-12-09
Do not include metadata annotations in Gemini's JSON schema


grighetto:
We've added metadata support in #33 but that should have been added only to the anonymized file when exporting the CQL statements. Instead, it was also included in the JSON output, which makes Gemini fail with:

```
Error: cannot create schema: invalid character '/' looking for beginning of value
```

┆Issue is synchronized with this [Jira Task](https://datastax-oss.atlassian.net/browse/AD-61) by [Unito](https://www.unito.io)
┆Issue Number: AD-61

PR #73, opened 2020-12-09
#72 Remove metadata from JSON export

absurdfarce:


python/adelphi/adelphi/store.py
```diff
@@ -49,7 +49,6 @@ def build_auth_provider(username = None,password = None):
 def with_cluster(cluster_fn, hosts, port, username = None, password = None):
     ep = ExecutionProfile(load_balancing_policy=default_lbp_factory())
     cluster = Cluster(hosts, port=port, auth_provider=build_auth_provider(username,password), execution_profiles={EXEC_PROFILE_DEFAULT: ep})
-    log.info("Connecting to the cluster to get metadata...")
```
absurdfarce, line 52: A minor nit I noticed while testing: this log message is completely out-of-date now

grighetto (approved):

Issue #48, opened 2020-11-12
Publish Python package in PyPI


grighetto:
We need to create a `setup.py` file for the anonymizer and distribute it as a package in pypi.org.
For some inspiration, we can follow the Python driver example: https://github.com/datastax/python-driver/blob/master/setup.py#L416

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1198513180644180/1198830660243597) by [Unito](https://www.unito.io/learn-more)

absurdfarce:
Note that this work will likely be folded into #40

grighetto:
@absurdfarce and I started this effort in different branches and ended up converging in what's currently in master.
Repurposing this issue to the actual publishing of the package in PyPI.

absurdfarce:
This wound up not being completely covered by #40 so we'll be pursuing it as it's own effort

absurdfarce:
@grighetto + @jdonenine ... do we agree there's nothing left to do for this issue now that the package is live?

jdonenine:
Yup, I actually thought I had closed it, apparently, I only moved it in the project... fail 😞 

![image](https://user-images.githubusercontent.com/7901615/102932950-4be09480-446f-11eb-8c03-c398f362fd31.png)

Close away!

absurdfarce:
:ohyeah:

PR #75, opened 2020-12-11
#48 PyPI publishing

absurdfarce:


python/adelphi/setup.py
```diff
@@ -13,17 +13,38 @@
...
+      'Natural Language :: English',
+      'Operating System :: OS Independent',
+      'Programming Language :: Python',
+      'Programming Language :: Python :: 2.7',
+      'Programming Language :: Python :: 3.5',
+      'Programming Language :: Python :: 3.6',
+      'Programming Language :: Python :: 3.7',
+      'Programming Language :: Python :: 3.8',
+      'Programming Language :: Python :: 3.9',
+      'Programming Language :: Python :: Implementation :: CPython',
```
absurdfarce, line 44: python-driver includes pypy support here.  We're _probably_ fine on pypy but I'm not explicitly testing on it so leave it off for now

python/adelphi/setup.py
```diff
@@ -13,17 +13,39 @@
...
+        'Natural Language :: English',
+        'Operating System :: OS Independent',
+        'Programming Language :: Python',
+        'Programming Language :: Python :: 2.7',
+        'Programming Language :: Python :: 3.5',
+        'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
+        'Programming Language :: Python :: 3.8',
+        'Programming Language :: Python :: 3.9',
+        'Programming Language :: Python :: Implementation :: CPython',
```
absurdfarce, line 45: python-driver includes pypy support here.  We _should_ be fine on pypy but I haven't tested it explicitly (and didn't want to introduce that as an ongoing thing) so for now I just dropped it.

python/adelphi/setup.py
```diff
@@ -13,17 +13,39 @@
...
+    name='adelphi',
+    url='https://github.com/datastax/adelphi',
+    description='Tooling for Adelphi',
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+    version='0.1.0',
+    project_urls={
+        'Source': 'https://github.com/datastax/adelphi',
+        'Issues': 'https://github.com/datastax/adelphi/issues',
+    },
```
absurdfarce, line 30: python-driver includes a "Documentation" link here which presumably shows up in the "Project links/Documentation" section on PyPI.  This documentation would presumably be specific to the Python package so we could make this a link to the README on Github... not sure what else would work there.

absurdfarce:
@grighetto and @jdonenine : Note that this PR is branched off of #71  rather than master.  I kinda doubted there would be any overlap between the two but just in case I decided to work off the more recent feature branch.  This PR will be rebased once #71 is merged to master so please don't merge this guy until we reach that point!

python/adelphi/README.md
```diff
@@ -0,0 +1,107 @@
+# Adelphi tooling
+
```
jdonenine, line 2: What do you think about adding a quick intro block to explain the high level of the script and what all it can do?  That will give a bit more context to the detailed commands you cover below, I think.

jdonenine: Maybe it can include some info on how to install the package via pip and what the basic system requirements are (python version, anything else)?

python/adelphi/README.md
```diff
@@ -0,0 +1,107 @@
...
+## Commands
+
+### export-cql
+This command extracts schemas for the specified keyspaces from a Cassandra instance and then displays the CQL commands necesary to generate them to standard out.  You can optionally specify an output directory, in which case the CQL commands are written to files within that directory, one file for each keyspace.
```
jdonenine, line 6: If we change the output structure so that we get a directory and a series of files plus the metadata file we'll need to remember to come back here and update this.

jdonenine: Based on the help dump below, I think this command is optional, and if it's left out you get all keyspaces by default - can we include a note about that here?

jdonenine: Actually, I guess looking at the other blocks here that default behavior is more for the --keyspaces option than it is the individual commands, so no need to repeat it over and over again.  But, it makes me wonder if we shouldn't pull the options section up higher and then expand on the detailed commands afterwards maybe?

python/adelphi/README.md
```diff
@@ -0,0 +1,107 @@
...
+      --username TEXT               Database username
+      --password TEXT               Database password
+      --keyspaces TEXT              Comma-separated list of keyspaces to include.
+                                    If not specified all non-system keypaces will
+                                    be included
+    
+      --rf INTEGER                  Replication factor to override original
+                                    setting. Optional.
+    
+      --anonymize / --no-anonymize  Enable/disable schema anonymization  [default:
```
jdonenine, line 67: Here I don't think that noting the default as True makes it clear which is the default since it's sort of an either/or situation, not a single flag with a boolean variable.

python/adelphi/setup.py
jdonenine: If we can link back to our docs in the readme that seems like the best scenario so that we only have to manage it in one place over time.

jdonenine (changes requested):

python/adelphi/README.md
```diff
@@ -0,0 +1,107 @@
...
+### export-cql
+This command extracts schemas for the specified keyspaces from a Cassandra instance and then displays the CQL commands necesary to generate them to standard out.  You can optionally specify an output directory, in which case the CQL commands are written to files within that directory, one file for each keyspace.
+
+The following will display the schemas for the keyspaces "foo" and "bar" on standard out: 
+
+    adelphi --keyspaces foo,bar export-cql
+
+If you wish to store the schemas in a directory "baz" you could use the following instead:
+
+    adelphi --keyspaces foo,bar --output-dir=baz export-cql
```
grighetto, line 14: `keyspaces` has no `=` and output-dir` has it, I think that could be confusing. I tend do prefer having the equals sign, but I don't care much either way as long we're consistent.

python/adelphi/README.md
grighetto: typo: necesary

python/adelphi/README.md
```diff
@@ -0,0 +1,107 @@
...
+To display Gemini-formatted schemas for the keyspaces "foo" and "bar" use the following:
+
+    adelphi --keyspaces foo,bar export-gemini
+
+And to store these schemas in a directory "baz":
+
+    adelphi --keyspaces foo,bar --output-dir=baz export-gemini
+
+### contribute
+This command automates the workflow of contributing one or more schemas to the Adelphi project.  The [Adelphi schema repository](https://github.com/datastax/adelphi-schemas) is implemented as a Github repository and contributions to this repository take the form of pull requests.  The workflow implemented by this command includes the following steps:
```
grighetto, line 28: maybe we can call out here that the `schemas will be *publicly* available in anonymized form` (or something along those lines), such that no one has any surprises

python/adelphi/README.md
```diff
@@ -0,0 +1,107 @@
...
+    
+    Commands:
+      contribute     Contribute schemas to Adelphi
+      export-cql     Export a schema as raw CQL statements
+      export-gemini  Export a schema in a format suitable for use with the the...
+
+Individual commands may have their own options and/or help text.  For example the help for the "contribute" command is as follows:
+
+    $ adelphi contribute --help  
+    Usage: adelphi contribute [OPTIONS]
```
grighetto, line 90: that `click` lib makes this a little weird, sometimes options come before, other times after. I understand it gives some positional context to the args, but I'd prefer to pass the `--aaa`  options at whatever position and it would still handle it. Not related to this PR of course, but we can probably check later if we can be more flexible here.

python/adelphi/setup.py
absurdfarce: I wondered about that @jdonenine, but I held off since we were already linking to lots of stuff on Github.  cassandra-driver links to some DataStax-managed docs but I think that might just be a historical accident; I'm not sure that's the model we should aim for going forward.

I'll make this a link to the README for now, although weirdly it'll be a link to the README being added by this very pull request.  Remember that the docs we link to here should be specific to the Python package in question so we don't want to link to https://github.com/datastax/adelphi/blob/master/README.md.... I think we'd want https://github.com/datastax/adelphi/blob/master/python/adelphi/README.md instead.

python/adelphi/README.md
absurdfarce: Yeah, I think that makes sense @jdonenine.  It seemed weird to add that kind of thing to a doc that's appearing on PyPI where presumably users are either already aware of those details of they're specified via classifiers and other metadata in setup.py.  But the README here is serving two distinct roles; it's providing content for PyPI _and_ it's serving as a general README in the Github repo.  The block you're describing makes a ton of sense for the second role so I'll add that in there.

python/adelphi/README.md
absurdfarce: Your comments did highlight a couple things that are missing here @jdonenine :

* If keyspaces aren't specified all keyspaces are selected, at least initially
* System keyspaces are removed from the list whether you explicitly specify keyspaces or leave them out and select everything

I'll add comments about both those points to the docs.

python/adelphi/README.md
absurdfarce: Fair point.  This is a weirdness of how the Python CLI arg lib we're using handles things.  When you declare a boolean flag it adds two CLI args, one for the name and one as "no-" name.  The flag is then turned on or off based on which one you add, so there is something of a decoupling there.

I think I'll change the default to something like "anonymization is enabled" to make that clearer.

python/adelphi/README.md
absurdfarce: Good catch @grighetto , will fix in the next push

python/adelphi/README.md
absurdfarce: I see your point... kinda.  I personally like the fact that the position of the args is used to be very explicit about what is command-specific and what is just used as part of running the app.

I _believe_ click supports the ability for command declarations to "inherit" options from their parent container (or something along those lines... I forget exactly how it works now).  Upshot is that all the args which are currently not tied to a command would devolve into command-specific args, which I believe would accomplish what you're describing.  I guess I'm not completely opposed to that; I even experimented with something like that when click first was brought in.  I personally still like the clear demarcation of arg types but I readily agree this may be an idiosyncrasy about me. :)

python/adelphi/setup.py
jdonenine: 👍  on it needing to be the python script specific README, not the general repo one.

python/adelphi/README.md
jdonenine: Agree on people getting here likely by knowing what they were looking for, so not a lot of detail necessary I'd think.

absurdfarce: I tried to address this in the "Introduction" section.  Take a look and see if you're satisfied with what's there @jdonenine

python/adelphi/README.md
absurdfarce: I wound up adding a section to the "Options" block describing how keyspace selection works.  I think it covers all the high points we discussed here.  Can you guys double-check and confirm that everything that needs to be there is there and that it's clear?

absurdfarce: I wound up adding a section to the "Options" block describing how keyspace selection works.  I think it covers all the high points we discussed here.  Can you guys double-check and confirm that everything that needs to be there is there and that it's clear?

python/adelphi/README.md
absurdfarce: This should be fixed now; I added an explicit "=" for all examples in the docs

python/adelphi/README.md
absurdfarce: This concept made it's way into the "Introduction"; it seemed important enough to warrant an explanation right up front.  Please take a look and let me know if that addresses the concern here.

python/adelphi/README.md
absurdfarce: Okay, this now looks like...

```
...
  --anonymize / --no-anonymize  Enable/disable schema anonymization  [default:
                                anonymization enabled]
...
```

To me that's clearer all around.  Whaddya think @jdonenine ?

python/adelphi/README.md
```diff
@@ -0,0 +1,129 @@
...
+      --token TEXT  Personal access token for Github user
+      --help        Show this message and exit.
+
+### A quick note on keyspaces
+None of the commands above *require* you to specify keyspaces for export.  If you do not supply the "--keyspaces" argument then *all* keyspaces will be considered for export.  In either case the application will prune system keyspaces before performing the export.
+
+### A quick note on anonymization
+The anonymization process can be explicitly enabled or disabled using the "--anonymize" and "--no-anonymize" arguments (respectively).  The default value will anonymize schemas so unless you explicitly wish to disable anonymization you don't need to supply either argument.  Also note that since all contributed schemas *must* be anonymized the "--no-anonymize" argument cannot be used when contributing schemas to Adelphi.  Supplying this argument when contributing one or more schemas will cause the application to exit with an error message.
+
+### Paramters via environment variables
```
grighetto, line 123: typo: paramters

python/adelphi/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+* Extraction of schemas for one or more keyspaces from a running Cassandra cluster
+* Optionally anonymizing these schemas
+* Formatting these schemas as CQL statements or as JSON documents
+* Displaying these formatted schemas on standard out or writing them to the filesystem
+* Automate a workflow for contributing anonymized schemas to the public [Adelphi schema repository](https://github.com/datastax/adelphi-schemas)
+
+The anonymization process replaces all keyspace names, table names and table column names with a generic identifier.  You can use the "adelphi" application to extract, format and display schemas from your Cassandra clusters without contributing these schemas to the Adelphi project, and for this use case anonymization is not required.  Anonymization *is* required anytime you contribute schemas to the Adelphi project.  All the schemas in our repository are publicly visible so to avoid *any* possible leakage of proprietary information we can only accept schemas which have been anonymized.
```
grighetto, line 12: I'd highlight the last part about visibility with some code-block style or at least move it to a new paragraph.

grighetto (approved):
Left a minor suggestion and found a typo, otherwise 👍

python/adelphi/README.md
absurdfarce: Good find, fixed!

python/adelphi/README.md
absurdfarce: Hmmm, the content of that line fits in with the rest of that paragraph (in fact that paragraph was structured so that it would lead up to that statement) so I don't really want to decouple them.  But I think perhaps making just that sentence bold would satisfy both of us.  I'll give that a shot.

absurdfarce:
I fixed the typo you identified @grighetto and was able to come up with a way to emphasize the public nature of the schema repo that I think will make both of us happy.  Would you mind taking a quick look to confirm that you're good?

python/adelphi/README.md
jdonenine: I like it, thanks for adding.

jdonenine (approved):
Good from my perspective, I'll let @grighetto wrap up on the couple of comments he originally had and finish things up.  Thanks for adding the docs!

grighetto (approved):
LGTM +1

Issue #74, opened 2020-12-11
Schema metadata should be friendly to automated parsing


absurdfarce:
Current format for metadata stored in schemas contributed by the "adelphi" tool is a homegrown format designed for ease of reading... but it isn't super-easy to parse in an automated fashion:

```
//Schema metadata:
//@created_at = 2020-12-10T09:52:39.793838
//@github_user = absurdfarce
//@hashed_ks_name = 9oeoJFuUPUpGOIjokdIIuwsJcmGny03H8MyjkovpV4w=
//@dc-count = 1
//@host-count = 1
```

This should be converted to something more friendly to programmatic parsing

absurdfarce:
Another option here would be to decouple schema and metadata.  Create a separate metadata file using some well-known name or in some well-known location containing metadata for each schema.  The schema file can then be pure content and the metadata file could be something very parse-friendly: JSON, TOML, whatever.

Think how Maven or Gradle structure their package repositories, something like that

jdonenine:
That approach would definitely solve the formating question.  We could go with some straightforward like json or yaml in a plain file.  I'd think the challenge might be how to preserve the linkage between the two files.

Maybe create a directory for each schema and put the two side by side, the metadata could contain a reference to the schema file path perhaps?

user -> 
  hashedschemaid -> 
    hashedschemaid
    hashedschemaid.metadata.json

Something like that? What are you thinking?

absurdfarce:
I haven't formed a really good opinion yet but I was considering something along the lines of what you're describing @jdonenine .  I was thinking a directory per schema; that directory would contain the raw schema file and the metadata file as discrete files.  The schema file would just be a straight text file (containing CQL/Gemini depending on what was selected for export) while the metadata would be... probably JSON to keep it simple.  We can just use well-known names for the files within the schema directory... something like "schema.txt" and "metadata.json" should work.

The big advantage I see to something like that is it provides a clean avenue for future expansion.  By keeping schema-specific files in their own directory you create isolated spaces where you can do interesting things going forward.  So say at some point you want to track versions of an individual schema over time you could do so; you'd need "schema-timestamp.txt" for a resource name, but you see where I'm going with this.

PR #77, opened 2020-12-16
#76 Python 3 incompatibility in export_gemini

jdonenine:

Issue #62, opened 2020-11-25
Convert integration test run script to Python, integrate with test frameworks for more granular feedback


absurdfarce:
Current work on the anonymizer test suite (see https://github.com/datastax/adelphi/pull/54) is adding an integration test suite which spins up various versions of C* via Docker, loads them up with a complex schema and performs anonymization on the resulting keyspace(s).  The driver for this effort is currently implemented as a shell script with results returned via the return value... but this only allows a binary fail/pass determination to be made.

Ideally we'd like to get individual feedback of success or failure for each C* version.  A straightforward way to accomplish this would be to conver this runner to Python and bring in a well-known testing framework.

┆Issue is synchronized with this [Jira Task](https://datastax-oss.atlassian.net/browse/AD-105) by [Unito](https://www.unito.io)
┆Issue Number: AD-105

PR #78, opened 2020-12-17
#74 Metadata reform

absurdfarce:
Moving away from metadata + schema in a single file and towards a directory containing all materials for a given schema with raw schema output (in either CQL or Gemini format) in one file and metadata (in JSON) stored in a separate file underneath that directory.

### DIFF
--- a/python/adelphi/adelphi/export.py
+++ b/python/adelphi/adelphi/export.py
@@ -89,7 +89,8 @@ class BaseExporter:
     def get_cluster_metadata(self, cluster):
         hosts = cluster.metadata.all_hosts()
         unique_dcs = self.__unique((host.datacenter for host in hosts))
-        return {"host-count": len(hosts), "dc-count": sum(1 for i in unique_dcs)}
+        unique_cass_vers = self.__unique((host.release_version for host in hosts))
+        return {"host-count": len(hosts), "dc-count": sum(1 for _ in unique_dcs), "cassandra-versions": ",".join(unique_cass_vers)}
 
 
     def get_common_metadata(self, cluster, props):
--- a/python/adelphi/adelphi/export.py
+++ b/python/adelphi/adelphi/export.py
@@ -1,0 +1,100 @@
+# Copyright DataStax, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import hashlib
+import logging
+from base64 import urlsafe_b64encode
+from datetime import datetime, tzinfo, timedelta
+
+try:
+    from itertools import ifilterfalse as filterfalse
+except ImportError:
+    from itertools import filterfalse
+
+from adelphi import NoKeyspacesSelectedException
+from adelphi.anonymize import anonymize_keyspace
+from adelphi.store import build_keyspace_objects
+
+
+log = logging.getLogger('adelphi')
+
+try:
+    from datetime import timezone
+    utc = timezone.utc
+except ImportError:
+    class UTC(tzinfo):
+        def utcoffset(self, dt):
+            return timedelta(0)
+        def tzname(self, dt):
+            return "UTC"
+        def dst(self, dt):
+            return timedelta(0)
+    utc = UTC()
+
+class BaseExporter:
+    
+    # unique_everseen from itertools recipes
+    def __unique(self, iterable, key=None):
+        "List unique elements, preserving order. Remember all elements ever seen."
+        # unique_everseen('AAAABBBCCDAABBB') --> A B C D
+        # unique_everseen('ABBCcAD', str.lower) --> A B C D
+        seen = set()
+        seen_add = seen.add
+        if key is None:
+            for element in filterfalse(seen.__contains__, iterable):
+                seen_add(element)
+                yield element
+        else:
+            for element in iterable:
+                k = key(element)
+                if k not in seen:
+                    seen_add(k)
+                    yield element
+
+
+    def build_keyspace_id(self, ks):
+        m = hashlib.sha256()
+        m.update(ks.name.encode("utf-8"))
+        # Leverage the urlsafe base64 encoding defined in RFC 4648, section 5 to provide an ID which can
+        # safely be used for filenames as well
+        return urlsafe_b64encode(m.digest()).decode('ascii')
+
+
+    def get_keyspaces(self, cluster, props):
+        keyspace_names = props["keyspace-names"]
+        metadata = cluster.metadata
+        keyspaces = build_keyspace_objects(keyspace_names, metadata)
+
+        if len(keyspaces) == 0:
+            raise NoKeyspacesSelectedException
+
+        log.info("Processing the following keyspaces: %s", ','.join((ks.name for ks in keyspaces)))
+
+        # anonymize_keyspace mutates keyspace state so we must trap keyspace_id before we (possibly) call it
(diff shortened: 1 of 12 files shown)


absurdfarce:
Biggest change here is the conversion to "exporter" objects which can collect, store and provide access to keyspace objects and metadata information.  These objects replace the collection of maps we were passing around within the old bin/adelphi script.  This allows us to continue to move things out of bin/adelphi, freeing it to focus mostly on arg processing (via click) and a high-level workflow.  Github support is the notable exception there as big chunks of GH functionality are still inline in bin/adelpi.

Also note that this approach addresses the concerns that were being discussed in https://github.com/datastax/adelphi/pull/77#discussion_r544407997.  We no longer have separation between the gatherer of keyspace obects and the consumer of same; exporters are free to gather this info in whatever way suits them.  The example motivating that conversation is useful here: the state maintained by the CQL exporter is the {keyspace:keyspace_id} map we all know and love while the Gemini exporter just maintains a keyspace + keyspace_id.

absurdfarce:
Current status: most things should be working now, although I certainly haven't tested all the various combinations of params.  There's probably at least a pass or two of cleanup yet to be done.  And I haven't messed with the integration tests yet; they're almost certainly borked due to the output format changes.

python/adelphi/adelphi/export.py
```diff
@@ -0,0 +1,100 @@
...
+                yield element
+        else:
+            for element in iterable:
+                k = key(element)
+                if k not in seen:
+                    seen_add(k)
+                    yield element
+
+
+    def build_keyspace_id(self, ks):
```
absurdfarce, line 66: Currently only called from this class but this fn will almost certainly need to be available for testing.... so I just left it as it stands.

python/adelphi/adelphi/export.py
```diff
@@ -0,0 +1,100 @@
...
+        def tzname(self, dt):
+            return "UTC"
+        def dst(self, dt):
+            return timedelta(0)
+    utc = UTC()
+
+class BaseExporter:
+    
+    # unique_everseen from itertools recipes
+    def __unique(self, iterable, key=None):
```
absurdfarce, line 48: There was no plausible reason to make this visible beyond the use here in get_cluster_metadata

python/adelphi/adelphi/gemini.py
```diff
@@ -19,97 +19,128 @@
...
+            self.keyspace = ks
+            self.keyspace_id = keyspace_id
+
+
+    def each_keyspace(self, ks_fn):
+        ks_fn(self.keyspace, self.keyspace_id)
+
+
+    def export_metadata(self):
+        return {k : self.metadata[k] for k in self.metadata.keys() if self.metadata[k]}
```
absurdfarce, line 50: We can have metadata for Gemini schemas now!

grighetto (approved):
Manually tested with Python 2 and 3. Verified the unit and integration tests. Only failures were around the new metadata format, which is expected since that wasn't contemplated.

absurdfarce:
Integration tests should be fixed now.  I'd still like to do a bit more testing on this guy but I think we're getting pretty close to the point where we could take another pass at a review of this guy.

python/adelphi/bin/adelphi
```diff
@@ -131,106 +109,83 @@ def export(ctx, hosts, port, username, password, keyspaces, rf, no_anonymize, ou
...
-                output_file.write(export_fn(iter([keyspace_obj]), metadata, options))
-                del metadata['keyspace_id']
+                metadata_file.write(json.dumps(metadata))
+
+        exporter.each_keyspace(ks_fn)
 
 
 # ============================ Command implementations ============================
 @export.command()
+@click.option('--no-metadata', help="Disable display of metadata when writing to standard out", is_flag=True)
```
absurdfarce, line 155: Added mainly to make the integration test fixes work.  We're adding "creation_timestamp" as a metadata field so unless we want to pull apart the metadata as JSON in the integration test suite we need the ability to ignore timestamps (which will never match a canonical schema).  Making something like jq a part of the test integration framework seemed like overkill here... and all this will be done differently once this framework is moved to Python anyway.

absurdfarce: I decided _not_ to document this for now.  We already provide an example of how to get help for a specific command in the README so my working assumption was that going from that to "adelphi export-cql --help" wasn't a huge leap.  It's not _abundantly_ clear that this is an option we want to hype up anyway; it certainly could have utility for some users in some cases but I'm not sure we want to encourage it too much.

I'm open to other opinions on this one.

python/adelphi/integration-tests/adelphi-docker-tests.sh
```diff
@@ -31,24 +29,41 @@ function create_schema() {
...
 
-function compare() {
-	local version=$1
-	echo "Comparing..."
-	diff schemas/$version.cql output/$version.cql
-	diffExitCode=$?
+function _do_diff() {
+        local schemaFile=$1
+	echo "Comparing schemas/${version}.cql to $schemaFile..."
+        diff -Z schemas/$version.cql $schemaFile
```
absurdfarce, line 43: I actually did see several cases where tests failed due to things like a missing newline at the end of one of the files fed into diff

absurdfarce:
In case it's useful I used the following list for testing:

```
adelphi --keyspaces=testkeyspace,spacecraft export-cql
adelphi --keyspaces=testkeyspace,spacecraft --no-anonymize export-cql
adelphi --keyspaces=testkeyspace export-gemini
adelphi --keyspaces=testkeyspace --no-anonymize export-gemini
adelphi --keyspaces=testkeyspace,spacecraft export-cql --no-metadata
adelphi --keyspaces=testkeyspace,spacecraft --no-anonymize export-cql --no-metadata

rm -rf foo/*; adelphi --keyspaces=testkeyspace,spacecraft --output-dir=foo export-cql
rm -rf foo/*; adelphi --keyspaces=testkeyspace,spacecraft --no-anonymize --output-dir=foo export-cql
rm -rf foo/*; adelphi --keyspaces=testkeyspace --output-dir=foo export-gemini
rm -rf foo/*; adelphi --keyspaces=testkeyspace --no-anonymize --output-dir=foo export-gemini
rm -rf foo/*; adelphi --keyspaces=testkeyspace,spacecraft --output-dir=foo export-cql --no-metadata
rm -rf foo/*; adelphi --keyspaces=testkeyspace,spacecraft --no-anonymize --output-dir=foo export-cql --no-metadata

adelphi --keyspaces=testkeyspace,system_views --maturity=dev contribute
```

python/adelphi/integration-tests/adelphi-docker-tests.sh
```diff
@@ -59,17 +74,26 @@ pip install ../../adelphi
...
 versions=(2.1.22 2.2.19 3.0.23 3.11.9 4.0-beta3)
+
+allPass=""
 for v in "${versions[@]}"
 do
 	start $v
 	create_schema $v
-	anonymize $v
+	export_schema $v
 	compare $v
```
grighetto, line 88: suggestion: call `compare` twice (once for each schema output type - stdout and file) then you can simplify your compare function implementation by checking a single file and a single return code at a time

absurdfarce: Hmmmm, interesting idea.  I'm not sure how much that actually buys me though.  I still have to put that cumbersome logic to gather two return values (one from each schema comparison) and check to make sure they're both zero before showing the "All comparisons matched..." message for a given C* version.  Doesn't the approach you're suggesting just move that logic to this do block rather than doing it in the compare function?

I kinda like where you're headed with this, but if at the core we're just moving around where that comparison of two return values happens I don't know that it helps much.

absurdfarce:
Just realized I hadn't run through the test suite above with Python 3.x.  Just did so with 3.9.0 and everything worked as expected.

I'm now fairly comfortable with the changes in this PR and @grighetto has reviewed and approved them as well so I'm considering merging this in order to close it out.  I'd like to continue the discussion around the bash script above to see if there's a cleaner way to manage the returned status from both test runs for a given C* instance but I don't think that discussion warrants holding up the entirety of the change.  There's also the fact that that shell script will hopefully be going away soon (see https://github.com/datastax/adelphi/issues/62) so it's prolly not worth breaking our backs over.

Still pondering this a bit, but my inclination right now is to merge and call it done (for now).

Issue #64, opened 2020-12-01
Write results in parallel and expose files through nginx


grighetto:
The storage volume for the results should be configurable such that the user can choose the most appropriate settings depending on his env (cloud or local).

So far, we've had the following challenges:

* ReadWriteMany volumes work on local k3d clusters but not on GKE;
* Standalone NFS server works on GKE but doesn't work on k3d (due to DNS issues in kubelet).

What we should do instead is exposing the relevant storage settings in values.yaml (we already expose `storageClass`) such that we can configure Adelphi to use AWS S3 or Google Cloud Storage.

┆Issue is synchronized with this [Jira Task](https://datastax-oss.atlassian.net/browse/AD-68) by [Unito](https://www.unito.io)
┆Issue Number: AD-68

grighetto:
The original goal of this ticket was to provide a way to write results in parallel to some external storage service (NFS) mainly to work around the ReadWriteMany limitation in GKE. It turns out that setting up an additional NFS server (such as Filestore) beforehand would add more complexity to the overall process and that approach wouldn't cover all bases (some users might choose to store the results in a different way).

Instead, the approach we went with is exposing the result through nginx such that the files can be easily download and moved anywhere else.

The way we work around the ReadWriteMany limitation in GKE is by having separate PVCs per job (gemini, nosqlbench, cassandra-diff). Since these PVCs can't be mounted simultaneously, we wait until the Argo workflow is done to start the results webserver. The results server has a separate lifecycle from the workflow, it will continue running even after the workflow exits such that we are able to download the data when everything is done. But once the Helm chart is uninstalled, then the results server will be terminated too.

grighetto:
I'll update the title of this ticket to reflect the new approach and  the actual work done here.
When we actually have the need for pluggable storage, we can revisit this.

PR #87, opened 2021-01-15
#64 Results storage simplified

grighetto:
The original intent of #64 was to make the storage configurable to work around a specific issue: being able to write the result files simultaneously from different pods and make it work across k8s envs (local or Cloud). So, the main idea was that for k3d the user could use local storage and on GKE he could provision an NFS server (Google Filestore).

Instead, the solution I'm proposition here achieves the same goals (parallel writes and cross-env compatibility) while being self-contained (no NFS or other external service required).

The way it works is that each job (gemini, cassandra-diff, nosqlbench) gets a dedicated PVC and in the end of the workflow these same PVCs are all mounted by the webserver (nginx) to expose the files to the client, which can then be download and stored somewhere else. The webserver is kept running after the Argo Workflow exits, but it will be terminated when the Helm chart is deleted.

### DIFF
--- a/helm/adelphi/templates/cassandra-diff-results.yaml
+++ b/helm/adelphi/templates/cassandra-diff-results.yaml
@@ -13,7 +13,7 @@ spec:
       command: ["sh", "-c", "-e"]
       args:
         - VERSION=$(wget -q -O - https://storage.googleapis.com/kubernetes-release/release/stable.txt 2> /dev/null);
-          OUTPUT=/results/cassandra-diff;
+          OUTPUT=/results;
           wget -q -O /usr/sbin/kubectl https://storage.googleapis.com/kubernetes-release/release/$VERSION/bin/linux/amd64/kubectl;
           pip install cassandra-driver;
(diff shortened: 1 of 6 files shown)


grighetto:
ping @absurdfarce - just a friendly reminder to review this PR when you have time

absurdfarce:
@grighetto ACKed.  Review is in-flight.

absurdfarce:
After sorting out a few minor issues (including a bone-headed mistake on my part) this worked out of the gate.  I was able to access output files via the nginx container for nosqlbench (source and target), gemini and cassandra-diff.  I also _very much_ applaud the simplification over the previous approach using NFS; this seems like a much more straightforward expression of what we're trying to achieve.

We do need to add some notes about how to use this, specifically something around the process of setting up port forwarding in order to enable local access to the Web server.  My understanding is that @jdonenine will be tackling this in future work.

absurdfarce (approved):
Worked as expected, and the simplification over the NFS approach is _very_ much appreciated.

PR #88, opened 2021-01-26
#36 Added example values.yaml for GKE.

grighetto:
The default values.yaml bundled in Adelphi will assume local execution with k3d and as such the `storageClassName` config will remain `local-path` and `clusterSize` set to 1.

In this PR we're providing an example config for running on GKE, which overrides the `storageClassName` to `standard` and `clusterSize` to 3.

In order to run with the new config file, add the `-f` flag to the Helm install command, like this:

```
helm install adelphi helm/adelphi -n cass-operator -f helm/adelphi/gke-values.yaml
```

### DIFF
--- a/helm/adelphi/gke-values.yaml
+++ b/helm/adelphi/gke-values.yaml
@@ -1,0 +1,7 @@
+############################ 
+# Config overrides for GKE 
+############################
+
+storageClassName: standard
+
+clusterSize: 3


absurdfarce:
Seems like we should add some docs indicating that something equivalent to what's described above must be employed when running on GKE.  Honestly I'm thinking we should update README.md to contain that info... it seems like the natural home for it (since it already has detailed commands for running Adelphi in k3d).

absurdfarce (approved):
Change looks good.  I do think we need some additional docs here as well to make users aware that this file exists, that they _have_ to use it when running on GKE and how they can do so.

grighetto:
We have a separate ticket to improve the documentation on the setup process, I'll leave a note there. The current README still looks good for the default case with k3d.

Issue #86, opened 2021-01-12
Generate workload YAML from CQL schema file


grighetto:
# Feature Request
### Description

As it stands right now, the NoSQLBench step of the workflow still requires an workload YAML to be provided, that is, it cannot automatically generate the workload from just the CQL schema itself. On the other hand, Gemini already does that for data integrity testing. We want to make NSB be on par with that such that we can do automatic performance testing as well.

### Describe the solution you'd like

Build a tool that takes a CQL schema file and transforms it into the desired output. We'll start with NoSQLBench, but the parsing and transformation pieces could be general enough such that we could reuse it with Harry or Gemini.

### Additional context

┆Issue is synchronized with this [Jira Task](https://datastax-oss.atlassian.net/browse/AD-53) by [Unito](https://www.unito.io)
┆Issue Number: AD-53

absurdfarce:
Making some notes for myself here based on some initial analysis

* Requires ability to generate nosqlbench bindings for every underlying CQL type
    * For each type we would want to support sequence and random (based on some distribution) selection
    * Sequence support required for each type for the case where it appears in a primary key
    * Could add some exclusions for particular types...
        * A sequence of inet addresses may make sense logically but might not be something we want to explicitly support
* Next step: analyze table metadata
    * Must generate data for each column included in primary key
    * At least one column in primary key must also support a sequence type (to guarantee uniqueness)
        * Could be chosen at random from collection of primary key columns
        * Could also include a bias towards some types... favor ints if possible, that kind of thing
* Output of this process so far: a collection of tuples
    * Column name
    * Column type
    * nosqlbench binding
* From these inputs we should be able to build the insert/update CQL statements using a template

absurdfarce:
A couple additional comments from subsequent analysis.

Base version of nosqlbench config is the version currently used in Adelphi; see https://github.com/datastax/adelphi/blob/2c586b62f0dbe1980a5d48d1f1a2be6649af7825/helm/adelphi/values.yaml#L57-L138.  The default usage within the Adelphi workflow should remove the schema creation blocks (since that will have already been performed earlier in the Argo workflow) and the verification blocks (since we're dealing with performance rather than integrity and we're backstopped by Gemini on integrity anyway).  A base config with these options removed appears to work without issue.

Population of sample data into the tables should fill data for every column for every row written.  The question then becomes what we want to do for the read/write loops in nosqlbench (main-read and main-write in the config).  A number of options were considered here:

* Just select all columns all the time
[...]

PR #96, opened 2021-02-03
#86 Base support for nb.yaml generation

absurdfarce:


python/adelphi/adelphi/nb.py
```diff
@@ -0,0 +1,127 @@
...
+        self.props["anonymize"] = False
+
+        self.metadata = self.get_common_metadata(cluster, self.props)
+
+        all_keyspaces = self.get_keyspaces(cluster, self.props)
+        if len(all_keyspaces) > 1:
+            raise TooManyKeyspacesException
+        (ks, keyspace_id) = next(iter(all_keyspaces.items()))
+        if (len(ks.tables)) > 1:
+            raise TooManyTablesException
```
absurdfarce, line 74: Only required because we want to constrain query generation to a single table and we don't currently support constraining operations by table at the "adelphi" tool level like we do with keyspaces (i.e. we don't have a "--table" arg for adelphi.bin.adelphi.export()).  We could add that functionality and avoid this check here (or more accurately convert it to an assert).... but I'm not sure how generally useful such a flag would be.

I suppose we could also add it as a flag specific to adelphi.bin.adelphi.export_nb()... that might make more sense.

python/adelphi/adelphi/nb.py
```diff
@@ -0,0 +1,127 @@
...
+    def binding_name(col):
+        return seq_binding_name(c) if col in primary_keys else dist_binding_name(c)
+    col_bindings = ",".join(["{" + binding_name(c) + "}" for c in cols])
+    return "insert into {}.{} ({}) values ({})".format(keyspace.name, table.name, col_names, col_bindings)
+
+
+class UnsupportedColumnTypeException(Exception):
+    """Exception indicinating that we can't create nosqlbench structures for a column type"""
+    def __init__(self, msg):
+        self.message = msg
```
absurdfarce, line 56: adelphi.exceptions defines exceptions common to all commands... it's also entirely fine for a command to define their own custom exceptions as necessary.

python/adelphi/adelphi/nb.py
```diff
@@ -0,0 +1,127 @@
...
+
+        # Always disable anonymization when generating nosqlbench configs
+        self.props = props.copy()
+        self.props["anonymize"] = False
+
+        self.metadata = self.get_common_metadata(cluster, self.props)
+
+        all_keyspaces = self.get_keyspaces(cluster, self.props)
+        if len(all_keyspaces) > 1:
+            raise TooManyKeyspacesException
```
absurdfarce, line 71: Seemed like a reasonable compromise for now.  We could create distinct nosqlbench configs for each input keyspace (along the lines of what we do for Gemini) but that felt pretty artificial.  And we don't yet have a clear understanding of what a nosqlbench config for multiple keyspaces looks like... do we want to generate select/insert statements for everything in _all_ the keyspaces in such a case?

Seemed safer to go with the minimal path for now.

python/adelphi/adelphi/export.py
```diff
@@ -98,4 +98,7 @@ def get_common_metadata(self, cluster, props):
         metadata.update(self.get_cluster_metadata(cluster))
         metadata["creation_timestamp"] = datetime.now(utc).isoformat()
         return metadata
-        
+
+    # Base impl... can be overridden if necessary
+    def each_keyspace(self, ks_fn):
+        ks_fn(self.keyspace, self.keyspace_id)
```
absurdfarce, line 104: This change (plus the removal of this explicit default impl from the subclasses who are now just using the superclass default) isn't really related to nosqlbench support... it's just a thing I noticed while working on the code.

python/adelphi/adelphi/nb.py
```diff
@@ -0,0 +1,132 @@
...
+        root["scenarios"] = {"default":[rampup_scenario, main_scenario]}
+        
+        root["bindings"] = build_bindings(self.table)
+
+        cl_map = {"cl":"LOCAL_QUORUM"}
+        cl_ratio_map = {"ratio":5}
+        cl_ratio_map.update(cl_map)
+
+        rampup_block = {"name":"rampup", "tags":{"phase":"rampup"}, "params":cl_map, "statements": [self.__build_rampup_statement()]}
+        main_read_block = {"name":"main-read", "tags":{"phase":"main", "type":"read"}, "params":cl_ratio_map, "statements": [self.__build_main_read_statement()]}
```
grighetto, line 128: I'm seeing this for the params:

```
- name: main-read
  params: &id001
    cl: LOCAL_QUORUM
    ratio: 5
```

python/adelphi/adelphi/nb.py
```diff
@@ -0,0 +1,132 @@
...
+        main_write_block = {"name":"main-write", "tags":{"phase":"main", "type":"write"}, "params":cl_ratio_map, "statements": [self.__build_main_write_statement()]}
```
grighetto, line 129: and here:

```
- name: main-write
  params: *id001
  statements:
```

python/adelphi/adelphi/nb.py
```diff
@@ -0,0 +1,132 @@
...
+        if col_name in [c.name for c in table.primary_key]:
+            if col.cql_type not in SEQS.keys():
+                raise ColumnTypeException("No nosqlbench sequence configured for type {}".format(col.cql_type))
+            rv[seq_binding_name(col)] = SEQS[col.cql_type]
+
+    return rv
+
+
+def build_select_statements(keyspace, table):
+    key_bindings = " and ".join(["{} = {}".format(key.name, "{" + dist_binding_name(key) + "}") for key in table.primary_key])
```
grighetto, line 42: here we probably have to quote upper/mixed-case entity names (unless NoSQLBench handles it internally) otherwise C* will complain

grighetto (approved):
I'm on board with the general approach here 👍 
Only found a couple of bugs that would be good to address in this PR, otherwise +1

python/adelphi/adelphi/nb.py
absurdfarce: This represents the declaration of an anchor in the YAML document.  In essence this defines a "common" structure in the YAML which can be referenced from other locations.  https://en.wikipedia.org/wiki/YAML#Advanced_components provides some additional context.

I was surprised by this too when it showed up; PyYAML apparently just does this by default when you re-use the same Python struct in the Python tree.

python/adelphi/adelphi/nb.py
absurdfarce: This is a reference to the anchor we declared earlier.  The earlier Wikipedia citation has some additional info here as well.

All of this is legit YAML (it shows up in the official spec) and nosqlbench appears to run fine with this config so the YAML parser in play there seems okay with it.  I can do a bit of checking to see if this feature can be disabled for PyYAML but I don't think it's essential to do so.

python/adelphi/adelphi/nb.py
absurdfarce: Good point; I wondered a bit about that when I was putting this code together and forgot to follow up with it.  I'll do some more testing to see if nosqlbench does the right thing... although I'll prolly just wind up quoting it here anyway.

absurdfarce: Latest commit added explicit quoting for keyspace, table and column names in generated statements.  A run with nb appears to be okay.

Issue #94, opened 2021-02-03
Provide a kubernetes service for the results-server pod


jdonenine:
# Feature Request
### Description
At the end of the Adelphi test run, the results-server pod is left running which exposes the results and makes them available for download via HTTP.  It would be nice if a service was also deployed mapping to the pod, this would make it easier to deploy ingress rules to access the data from outside of the cluster.

### Describe the solution you'd like
Automatically deploy a kubernetes service to map traffic to the results-server pod.

### Describe alternatives you've considered
Manually deploy a service after the workflow has completed, prior to configuring any ingress rules.

`kubectl port-forward` is also a good solution for a local development cluster, but not ideal for a shared existing cluster where ingress is likely already present for external access.

### Additional context
Locally `kubectl port-forward pod/results-server --namespace cass-operator :8080` works well.

┆Issue is synchronized with this [Jira Task](https://datastax-oss.atlassian.net/browse/AD-48) by [Unito](https://www.unito.io)
┆Issue Number: AD-48

PR #97, opened 2021-02-03
Updates to README

jdonenine:


README.md
```diff
@@ -1,75 +1,328 @@
...
+
+### Kubernetes Environment
+
+Adelphi is built to run on your existing kubernetes infrastructure, if you don't already have access to a kubernetes environment, running a kubernetes cluster locally is a great way to get started.  K3d (discussed below) provides an easy way to run a kubernetes cluster within Docker.  K3d runs on top of Docker, so if you go this route, Docker will also be required.
+
+### k3d
+
+To get started with k3d, check out their [documentation](https://k3d.io/).
+
+Once kind is installed, start by deploying a cluster:
```
grighetto, line 60: once *k3d* is installed

README.md
```diff
@@ -1,75 +1,328 @@
...
 ```
 git clone https://github.com/datastax/adelphi
-cd adelphi
+```
+
+Or download it manually from [GitHub](https://github.com/datastax/adelphi/archive/master.zip) and extract the contents of the archive.
+
+### Deployment
+
+As mentioned, Adelph is deployed using helm.  Within the Adelphi repository you will find a series of helm charts in the `/helm` directory.
```
grighetto, line 129: first Adelphi mention is missing an i

grighetto: > Within the Adelphi repository you will find a series of helm charts in the `/helm` directory

Technically, it's a single chart with multiple templates/manifests.

README.md
```diff
@@ -1,75 +1,328 @@
...
+
+In general, with helm, there are two ways to modify the values provided to a deployment:
+* via the command line using `--set value.path=value`
+* via a `yaml` file via `-f custom.values.yaml`
+
+## Observing Workflow Progress
+
+### Argo CLI
+
+Once Adelphi has been deployed and the workflow has begun, the Argo CLI can be used to monitor it's progress:
```
grighetto, line 157: *its*

README.md
```diff
@@ -1,75 +1,328 @@
...
+
+```
+argo get -n cass-operator @latest
```
grighetto, line 160: I've never used `get`, does it report the progress once and exit?
I like `argo watch` because it keeps refreshing the output.

README.md
```diff
@@ -1,75 +1,328 @@
...
+        backend:
+          service:
+            name: argo-server
+            port:
+              number: 2746
+```
+
+You should now be able to lauch the Argo web application from a browser at `http://argo.{ADELPHI_LB_ADDRESS}.nip.io/` to visually inspect and monitor the testing workflow.
+
+> This example uses nip.io to provide wildcard DNS support.  To learn more about nip.io, check out their [documentation](https://nip.io/).
```
grighetto, line 248: > This approach may be necessary if you're running on an existing cluster or a cluster deployed within a cloud provider

FYI, port forwarding works just as well on GKE too.
It's great you added this additional information on ingress, but for a short-lived deployment such as Adelphi, I think port-forwarding works fine and it's a consistent, low effort, approach across local and cloud envs.
I'm wondering if we could move this part to the end as an Appendix.

README.md
```diff
@@ -1,75 +1,328 @@
...
+ ├─✔ nosqlbench-source         nosqlbench-job/nosqlbench-source                    workflow-adelphi-1-1304337547  1m                                      
+ ├─✔ nosqlbench-target         nosqlbench-job/nosqlbench-target                    workflow-adelphi-1-2207475351  1m                                      
+ ├─✔ run-diff                  spark-job/spark-job-template                        workflow-adelphi-1-2458661793  4m                                      
+ ├─✔ collect-diff-results      cassandra-diff/collect-results                      workflow-adelphi-1-1334654991  16s                                     
+ ├─✔ gemini-oracle-sut         gemini-job/gemini-oracle-sut                        workflow-adelphi-1-3173132751  2m                                      
+ └─✔ start-results-server      results/results-server                              workflow-adelphi-1-1122498803  2s  
+ ```
 
-## Pre-requisites
+> Note that the `build-cassandra` and `build-cassandra-mgmt-api` steps of the workflow may not be executed with all runs of the workflow.
```
grighetto, line 286: since we're mentioning it, it's probably good to let people know why: `unless you have specified a git_identifier to build C* from the source` (or something along those lines).

README.md
```diff
@@ -1,75 +1,328 @@
...
-    ```
+This will enable port forwading on your local machine on an available port:
 
-4. If you don't have Helm installed you can [install](https://helm.sh/docs/intro/install/) it with the following command:
+```
+Forwarding from 127.0.0.1:64058 -> 8080
+Forwarding from [::1]:64058 -> 8080
+```
+
+You should now be able to lauch the Adelphi results web server from a browser on your local machine at http://localhost:64058/ to download the various results files generated by Adelphi.
```
grighetto, line 318: > You should now be able to lauch the Adelphi results web server from a browser on your local machine at http://localhost:64058/

As an user, this sentence would make me think the server would always be available at that port, but it was selected randomly at runtime. Maybe we can either set a port mapping `8080:8080` in the example or call out to check the port in the console output (`replace with the port you see in the output above`)

jdonenine:
Thanks for all the feedback @grighetto, I'll work all of this into the changes I'm still working on and get them in today.  I'm thinking I will actually split the getting started guide out from the initial README, so I'll try to restructure around the ingress info too as you mentioned.

README.md
```diff
@@ -1,75 +1,328 @@
...
+
+### Configuration
+
+To control the deployment configuration, many options are made available via the helm charts included with Adelphi.
+
+To learn more about setting values via helm, see their [documentation](https://helm.sh/docs/chart_template_guide/values_files/).
+
+In general, with helm, there are two ways to modify the values provided to a deployment:
+* via the command line using `--set value.path=value`
+* via a `yaml` file via `-f custom.values.yaml`
```
grighetto, line 151: It's worth mentioning here that the default configuration is for running locally on k3d.
For running on GKE, Adelphi comes with a custom values YAML that can be used with:

```
helm install adelphi helm/adelphi -n cass-operator -f helm/adelphi/gke-values.yaml
```

README.md
jdonenine: Yeah, I went this route with the auto-selected port to avoid the problem of someone having a conflict on what we pick here and then being like, it doesn't work. The auto-selection will always work.  But, if you're not familiar, I can definitely see the confusion like you're mentioning.

grighetto: Also, it's probably worth mentioning here that you can download the results from the cluster with 

```
wget -r http://localhost:8080/
```

README.md
jdonenine: Yup, get is just a snapshot.  I wanted to include that so that I didn't have to explain how to escape the context it drops you into.

I'm adding a blurb about watch as well, so we'll have both.

README.md
jdonenine: I'm restructuring things so that in the core of the getting started flow we'll only discuss port-forward, and I've generalized that text some so it's not "local" specific.

I'll keep some of the content about Ingress in a new Appendix section.

So I think that will cover our bases here.  Although we'll need #94 to make ingress to the results-server easier.  But for now I can include an example of creating the service manually before exposing via ingress, that'll work for now.

README.md
jdonenine: 👍

README.md
jdonenine: I will definitely add that.  I didn't know until yesterday you mentioned it that it was possible, a good add for sure.

jdonenine:
@grighetto my latest couple of commits have the restructuring, which is going to make your re-review hard, because it's all moved, sorry about that.  But I weaved all of your previous comments into what is there now.  What I've generally done is split the Getting Started doc out from the base doc.

GETTING_STARTED.md
```diff
@@ -0,0 +1,333 @@
...
+
+Helm is a package manager for kubernetes. Adelphi packages a number of helm templates for managing the kubernetes deployment.  To learn more about helm and install it, check out the helm [documentation](https://helm.sh/).
+
+### Argo (recommended)
+
+Adelphi uses Argo to orchestrate the testing workflow within kubernetes.  The Argo CLI can be used to get info about the workflow and its progress once started.  To learn more about Argo and install the CLI tooling, check out the Argo [releases](https://github.com/argoproj/argo/releases).
+
+### Docker (recommended for local deployment)
+
+If you need to install a local Kubernetes cluster using at tool like k3d, as described later in this guide, you'll also need to have Docker running locally.  To learn more about Docker and install it, check out the Docker [documentation](https://docs.docker.com/get-docker/).
```
grighetto, line 23: > using at tool like k3d

using *a* tool

jdonenine: Bah, dang it, fixed, thanks!

GETTING_STARTED.md
```diff
@@ -0,0 +1,333 @@
...
+#### Storage Classes
+
+The default `storageClass` configured to be used by Adelphi is tailored for deployment via k3d, `local-path`.  This storage class may not be available on all clusters and should be modified as necessary.
+
+For example, when deploying to a cluster within [Google Kubernetes Engine](https://cloud.google.com/kubernetes-engine) it's necessary to customize the `storageClass` to be used.  An example of doing this is provided in [gke-values.yaml](helm/adelphi/gke-values.yaml) where `storageClass` is set to `standard`.
+
+This can also be applied via the helm CLI with:
+
+```
+helm install adelphi helm/adelphi --namespace cass-operator --set storageClass=standard
```
grighetto, line 135: I was actually hoping people would use the `gke-values.yaml` file directly with the `-f` param, such that they don't even have to worry about storage classes mean.
Another difference in that custom yaml is the cluster size, by default you have only one C* node per cluster, but with the custom gke values you'll have 3 (which is kind of required for real testing). So instead of setting yet another param in the command-line, we can just point to the file.

Anyway, no strong opinion here on the doc, but maybe we can at least add the alternative below:

```
helm install adelphi helm/adelphi -n cass-operator -f helm/adelphi/gke-values.yaml
```

jdonenine: Ah, OK, we can do that.

jdonenine:
Its worth mentioning that there's more I'd like to do here, but I think this is a good enough start to move forward with for the moment.  The other stuff I still want to do, but will do outside of this at some point here, next milestone I suppose:

* Basic architecture diagram - something visual
* Examples of more customized configurations - different versions, different basic settings type stuff
* Examples of using a real-world schema
* Additional Ingress - Adding an example for results-server in addition to the argo-server that's covered now

GETTING_STARTED.md
jdonenine: I added an example like you suggested, but did it up above in the basic deployment section, which makes it a bit more prominent.

grighetto (approved):
LGTM 👍 

I really liked the addition of the expected output for each command, so users can compare with what they get while following this guide.

Issue #102, opened 2021-02-04
Enhance testing for nosqlbench config generation


absurdfarce:
# Feature Request
### Description
We have a strong interest in making sure that the nosqlbench config generation process works well with various combinations of column types, primary/clustering key configurations, etc.  As a result we should implement a testing regime that maximizes this coverage.

This doesn't _necessarily_ mean that the tool has to handle every schema... but in situations where it doesn't we should make the failure very clear to the user.

### Describe the solution you'd like
It'd be nice if anything we do here could re-use as much of the testing infrastructure as possible

┆Issue is synchronized with this [Jira Task](https://datastax-oss.atlassian.net/browse/AD-42) by [Unito](https://www.unito.io)
┆Issue Number: AD-42

PR #104, opened 2021-02-23
62 testing is caring

absurdfarce:


absurdfarce:
Best approach to running these tests is probably to use tox (https://tox.readthedocs.io/en/latest/).  That's what was used when developing this test and looks to be a very useful framework for these tests going forward, especially for CI integration.  I've included a sample tox.ini file here... it works but it's really just a demonstration.

The included tox config tests against Python 2.7 and Python 3.6 with a single command using custom-build venvs.  You can specify only a single such version for testing using the -e flag; "tox -e py27" will just test the Python 2.7 deployment.  I've also added a number of environment variables which can be used to customize test operations:

* KEEP_CONTAINER = if set preserve the running Cassandra container in Docker (normally it's removed when the test wraps up).  Value doesn't matter; the env var just has to be set.
* KEEP_LOGS = if set preserve the temp directory used by the tests for storing files (normally it's removed when the test wraps up).  Value doesn't matter; the env var just has to be set.
* CASSANDRA_VERSIONS = a comma-delimited list of C* images to be tested.

python/adelphi/adelphi/export.py
```diff
@@ -81,9 +82,23 @@ def get_keyspaces(self, cluster, props):
...
+            tables = ks.tables.items()
+            ks.tables = OrderedDict(sorted(tables, key=lambda x: x[0]))
+            user_types = ks.user_types.items()
+            ks.user_types = OrderedDict(sorted(user_types, key=lambda x:x[0]))
+            indexes = ks.indexes.items()
+            ks.indexes = OrderedDict(sorted(indexes, key=lambda x:x[0]))
+
+            return ks
+
+        return {process_keyspace(ks) : self.build_keyspace_id(ks) for ks in keyspaces}
```
absurdfarce, line 101: While testing I started seeing diff failures based on ordering of tables within the keyspace.  The root cause here appeared to be a reliance on platform-specific functionality somewhere in the code combined with subtle behavioural differences on 2.7 vs. Python 3 in some weird cases.  I never did identify exactly what the platform-specific code was or why it was triggered in some cases but not others... in fact I could never reproduce it in practice.  But it consistently caused my tests to fail.

This implementation explicitly addresses the problem by enforcing an ordering based on name for tables, UDTs and indexes.  We do it after anonymization since that process might change those names.

python/adelphi/tests/unit/test_anonymize.py
```diff
@@ -6,7 +6,7 @@
 	FIELD_PREFIX,\
 	INDEX_PREFIX,\
 	get_name
-from schema_util import get_schema
+from tests.unit.schema_util import get_schema
```
absurdfarce, line 9: This + the addition of an __init__.py for tests were necessary to facilitate test discovery by pytest

absurdfarce:
Merged PR for https://github.com/datastax/adelphi/issues/102 into this PR so this change should be understood to apply to both issues.

python/adelphi/adelphi/nb.py
```diff
@@ -132,4 +132,4 @@ def __build_schema(self):
         main_write_block = {"name":"main-write", "tags":{"phase":"main", "type":"write"}, "params":cl_ratio_map, "statements": [self.__build_main_write_statement()]}
         root["blocks"] = [rampup_block, main_read_block, main_write_block]
 
-        return yaml.dump(root, default_flow_style=False)
+        return yaml.dump(root, default_flow_style=False).strip()
```
absurdfarce, line 135: I was seeing strange diff failures due to added or missing blank lines in diff output.  Underlying cause appeared to be differences in writing to stdout vs. whatever the Python YAML package does here.  Adding this strip() call seemed to bring everybody into agreement on the minimal case (i.e. the fewest number of blank lines possible).

python/adelphi/tests/integration/__init__.py
```diff
@@ -0,0 +1,161 @@
...
+        outputDirPath = self.outputDirPath(version)
+        os.mkdir(outputDirPath)
+        subprocess.run("adelphi --output-dir={} {} 2>> {}".format(outputDirPath, exportCmd, stderrPath), shell=True)
+        log.info("Adelphi completed")
+
+
+    def compareSchemas(self, version=None):
+        referencePath = self.getReferenceSchemaPath(version)
+        stdoutPath = self.stdoutPath(version)
+        rv1 = subprocess.run("diff -Z {} {}".format(stdoutPath, referencePath), shell=True)
```
grighetto, line 138: the `-Z` flag added in https://github.com/datastax/adelphi/commit/bd7c65294d5b5af35fb7aabadcc8ef465fdaf48e doesn't work on OSX, it's GNU specific.
These seem to be common options that we can use instead:

```
-b  --ignore-space-change
              Ignore changes in the amount of white space.

       -w  --ignore-all-space
              Ignore all white space.
```

At some point, we should probably move the diff'ing to Python too, maybe we could use something like this: https://docs.python.org/3/library/difflib.html

python/adelphi/tox.ini
```diff
@@ -0,0 +1,11 @@
+[tox]
+envlist = py27,py36
```
grighetto, line 2: It would be great if we could specify only the major versions here like `py2,p3`. Instead of py36, I only have py39 locally, so it failed the first time I tried to run the tests and I had to change this setting manually.

python/adelphi/tests/integration/__init__.py
absurdfarce: It's a good point... I'm prolly gonna wind up just taking the -Z out entirely.  I added it trying to address a mismatch I was seeing but I'm pretty sure I misread the mismatch; it was actually caused by blank _lines_ while -Z only ignores excess whitespace after an otherwise matching line.  I think -B is what I actually wanted and that's supported on OS X diff.

But none of that really matters because I completely agree with your other comment; the longer term goal is to move off of a reliance on shell commands.  I did some experimenting with difflib while working on this PR and ultimately decided against using it.  The format is somewhat non-standard (it looks vaguely diff-like but isn't exactly the same) and more importantly it only works when applied to two strings... meaning I'd have to read the entire contents of the file into memory in order to do a diff.  I... don't like that so much, so I was gonna look for a streaming solution... but that can wait.

So yeah, very much +1 to the idea of this being a short-term solution.

python/adelphi/tox.ini
```diff
@@ -0,0 +1,11 @@
...
+
+[testenv]
+deps = pytest
+       docker ~= 4.4.3
+       subprocess32 ~= 3.5.4
+setenv =
+       KEEP_LOGS=1
+       CASSANDRA_VERSIONS=4.0-beta3
```
grighetto, line 10: any reason not to include all versions here by default?

absurdfarce: Nah, this is a holdover from local testing... you're absolutely right, I just need to take this out

absurdfarce: I moved those settings over to a tox.ini.dev and removed them from the base tox.ini.

python/adelphi/tests/integration/test_diff.py
```diff
@@ -0,0 +1,23 @@
...
+    def setUp(self):
+        super(TestDiff, self).setUp()
+
+
+    def getExportCommand(self):
+        return "export-cql --no-metadata"
+
+
+    def getSchemaPath(self):
+        return "tests/integration/resources/diff-base-schema.cql"
```
grighetto, line 19: Since we're in Python land now, can't we call the function in `schema_util.py` to generate the base schema file on the fly instead of having to commit it as a text file?
Another benefit would be not having to remember to update that CQL file when changing the schema generator.

python/adelphi/tests/integration/resources/nb-base-schema.cql
```diff
@@ -0,0 +1,24 @@
+CREATE KEYSPACE testkeyspace WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '1'}  AND durable_writes = true;
```
grighetto, line 1: Can't we use the same base schema as the anonymization tests? We can probably ignore the unsupported data types for now (suppress them from the nb output).

absurdfarce: At the moment unsupported types break everything.  There are plans to fix this (see https://github.com/datastax/adelphi/issues/107 for additional details) but this has to be done carefully.  At minimum you have to make sure any primary keys are supported types (for reasons discussed in that ticket).  That ticket also envisions extending base type support to ints and (perhaps) date/time vals which should cover a lot of schemas... which I think includes the base schema as well?

I could see maybe revisiting this after https://github.com/datastax/adelphi/issues/107 is completed.  I really don't love the idea of maintaining a collection of static schema files for individual tests... but at the same time if that's the most efficient way to do it I don't want to close off that avenue.

So in short I guess I'm agreeing with you, but I'm not sure we have enough infrastructure to really do it right yet.  Fortunately I was planning on picking up https://github.com/datastax/adelphi/issues/107 as the next ticket... so maybe I'll take another look at this as part of that work.  Seem reasonable?

python/adelphi/tox.ini
absurdfarce: This is a tox thing, and I'm not sure how much control we have over it.  According to their docs (https://tox.readthedocs.io/en/latest/config.html) "py2" and "py3" should be supported.  Did you try those explicitly and see them fail?

I'll also give them a run locally and see if I can make them work.

One thing I did notice about tox; if you specify a Python version higher than the one you currently support you get errors about an inability to create a venv.  I have 3.6.x on my local box (that's where the "py36" came from) and when I tried to do a test with "py39" is complained about an inability to grab the interpreter.  So it's not like it's downloading Python versions or anything... this isn't rvm.  But when I make Python 3.9.x my active "py3" then "tox -e py36" seems to work okay.  It's... kinda strange; I definitely don't completely understand how tox is working yet.

I'll dig into this a bit more; I agree that if we can get away with "py2" and "py3" here that's better.  Well, we probably should require "py27" since we aren't supporting anything else in the 2.x line... but you see what I'm driving at.

grighetto: I just tried "py3" and it works 👍 
But "py2" doesn't, it fails with: `ERROR: InterpreterNotFound: python2`, but as you said, it's probably ok to have it as py27.

grighetto:
When running with python2, I'm seeing errors like this:

```
1c1
< CREATE KEYSPACE IF NOT EXISTS ks_1 WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '1'}  AND durable_writes = true;
---
> CREATE KEYSPACE IF NOT EXISTS ks_0 WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '1'}  AND durable_writes = true;
3c3
```

The keyspace id counter is off by one, I'm not sure if this was already an issue before.
It works with py3.

grighetto:
NB tests fail consistently with either python version:

```
tests/integration/__init__.py:74: in runTestForVersion
    self.runTestWithSchema(version)
tests/integration/__init__.py:154: in runTestWithSchema
    self.compareSchemas(version)
tests/integration/__init__.py:139: in compareSchemas
    self.assertEqual(rv1.returncode, 0)
E   AssertionError: 1 != 0
```

It seems it can't even connect to the cluster.

```
ERROR    cassandra.cluster:__init__.py:33 Control connection failed to connect, shutting down Cluster:
Traceback (most recent call last):
  File "cassandra/cluster.py", line 1690, in cassandra.cluster.Cluster.connect
  File "cassandra/cluster.py", line 3488, in cassandra.cluster.ControlConnection.connect
  File "cassandra/cluster.py", line 3533, in cassandra.cluster.ControlConnection._reconnect_internal
cassandra.cluster.NoHostAvailable: ('Unable to connect to any servers', {'127.0.0.1:9042': OperationTimedOut('errors=Timed out creating connection (5 seconds), last_host=None')})
```

python/adelphi/tests/integration/__init__.py
absurdfarce: Update: the diff dependency has been removed and comparisons on that front are made entirely in Python now

python/adelphi/adelphi/export.py
absurdfarce: Update: enforcement of explicit ordering within the *Metadata objects has been abandoned in favor of a new comparison strategy based on sha256 hashes for each CQL command in a given schema.

python/adelphi/tests/integration/test_diff.py
absurdfarce: Hmmmm, we might be able to do this for the diff test but not for the nosqlbench test (at least not yet) due to the brittleness of the current impl w.r.t. types.  Extending the schema_util impl to optionally add things based on params or a definition of some kind would probably solve that but I'm not sure that's something we need right now; the file-based approach does work in the short-term.

Maybe filing another ticket to expand the schema_util impl is the right way to go here.

I'll take a quick run at this and see how bad it is to change for the diff test.

Man, we also really ought to think about changing the name of that test... calling it a "diff" test is (a) wrong now and (b) telling us more about the impl of the test rather than what it's testing even if it were right.

python/adelphi/tests/util/schemadiff.py
```diff
@@ -0,0 +1,27 @@
...
+    for line in open(schemaFile):
+        realLine = line.strip()
+        if len(realLine) == 0 or realLine.isspace():
+            continue
+        buff += (" " if len(buff) > 0 else "")
+        buff += realLine
+        if realLine.endswith(';'):
+            cqlStmt = buff
+            buff = ""
+            m = hashlib.sha256()
```
absurdfarce, line 19: Fun fact I learned while working on this code: there's no way to reset a Python digest object (like the sha256 object we have here).  You just feed it data and when you call digest() it gives you the digest of whatever data it's seen thus far.  So if the schema we get from adelphi's "export-cql" is in a different order from what's in the reference schema you'll get digest mismatches for the same content because of the change in ordering.

And we know there are differences in ordering between those two files... that was one of the prime movers that led to checking things in Python via digests in the first place.

Creating a new digest for each complete statement and just using it immediately addressed the problem.  It strikes me as somewhat wasteful but it seems to be the Pythonic way for handling something like this.

python/adelphi/tests/integration/__init__.py
```diff
@@ -0,0 +1,192 @@
...
+                    continue
+                buff += (" " if len(buff) > 0 else "")
+                buff += realLine
+                if realLine.endswith(';'):
+                    log.debug("Executing statement {}".format(buff))
+                    try:
+                        session.execute(buff)
+                    except:
+                        log.info("Exception executing statement: {}".format(buff), exc_info=sys.exc_info()[0])
+                    buff = ""
```
absurdfarce, line 68: After recent changes this fn does now have some fairly strong resemblance to schemadiff.cqlDigestGenerator (see https://github.com/datastax/adelphi/pull/104/files#diff-2bc5a69b0a6a26da146863eadfe81e272b9cbf29ad2ee38992262a0d10ed8058R8-R23) but I think it still is worthwhile to keep this guy separate.  Doing so allows us to emphasize the role of cqlDigestGenerator() as a generator fn while thinking of this guy as a basic fn.

We certainly _could_ create a higher-level wrapper function which would contain the line-for-line reading from a file object.  We'd then have to supply another fn defining what do with each accumulated statement (i.e. the contents of a buffer once we find a trailng ';').  But note that these fns have different logic around when to ignore a line... so that would have to be abstracted too.  And at that point it seemed like it was getting complicated enough that just explicitly re-using the logic here was the cleanest bet.

python/adelphi/tests/integration/test_diff.py
absurdfarce: After https://github.com/datastax/adelphi/pull/104/commits/28232ade777bb9af2863c311aa148cf08c9248c2 and https://github.com/datastax/adelphi/pull/104/commits/808e4b00798276b898ef10da80630507a5e64d89 this should be working now.

python/adelphi/tox.ini
grighetto: This is still an issue for me. From a clean checkout, it fails because I don't have py36. I have to manually set it to py3.
Does it work with py3 for you? If so, any objections with going with that by default?

grighetto:
@absurdfarce I'm not seeing the test failures anymore, so I'm good with this PR.
The only outstanding issue is around py36 vs py3 in the settings. Is that something we can fix before merging?

python/adelphi/tox.ini
absurdfarce: Yeah, "py3" should work just fine... and should generalize better.  I'll remove the explicit minor version designation.

absurdfarce:
@grighetto There was an issue with the earlier implementation which was causing exceptions in a few spots to not result in test failures.  We were getting such an exception during creation of the schema on the DB (well before we actually do any compares) and the process was exiting out but, per discussion above, no test failure. #sadness

I've fixed this now, but with the current impl (as of this writing, so up to and including commit 579458f) there's still some kind of weird issue.  The schema for the diff test being generated by the Python code seems to be... wrong.  I'm getting the following:

```
InvalidRequest: Error from server: code=2200 [Invalid query] message="Invalid non-frozen collection type map<int, int> for PRIMARY KEY column 'my_column_10'" 
```

which makes complete sense given the schema that we're trying to add:

```
cqlsh> CREATE TABLE my_ks_0.my_table_1 (
   ...     my_column_10 map<int,int>,
   ...     my_column_11 set<int>,
...
   ...     PRIMARY KEY ((my_column_10, my_column_11), my_column_12, my_column_13)

```

What's weird is that that is _not_ the output I get when I just execute tests/util/schema_util.py:

```
(python3-venv) [mersault@linux python]$ python adelphi/tests/util/schema_util.py
...
CREATE TABLE my_ks_0.my_table_1 (                                                                                                                                                 
    my_column_0 ascii,                                                                                                                                                            
    my_column_1 bigint,
...                                                                                                                                              
    PRIMARY KEY ((my_column_0, my_column_1), my_column_2, my_column_3)  
```

I'm not sure where the disconnect is coming from yet.  The logic for writing the schema to the temp directory is identical to what's done when executing schema_util.py so they _shouldn't_ be any different.  I'm still looking into this and will update when it's resolved.

absurdfarce:
@grighetto The failing test issue discussed above was resolved by https://github.com/datastax/adelphi/pull/104/commits/4ec90d761cdd9c2cd949572a99141c34ac89e247.  The root cause was actually in schema_util; it was making some incorrect assumptions about the ordering of columns in the metadata objects and that was causing the random assignment of unexpected types to the partition and clustering columns of schemas created by that class.  I fixed the bugs and updated the anonymization tests (which were borked by the underlying bug).

As of this commit tox is clean for both Python 2 and 3.

python/adelphi/tests/integration/__init__.py
```diff
@@ -0,0 +1,200 @@
...
+                    continue
+                buff += (" " if len(buff) > 0 else "")
+                buff += realLine
+                if realLine.endswith(';'):
+                    log.debug("Executing statement {}".format(buff))
+                    try:
+                        session.execute(buff)
+                    except:
+                        log.error("Exception executing statement: {}".format(buff), exc_info=sys.exc_info()[0])
+                        self.fail("Exception executing statement: {}, check log for details".format(buff))
```
absurdfarce, line 106: Failures to create the test schema will now fail the test.  This is more rigorous and seems more correct on it's face but note that we _were_ using this operation to silently handle the case of an inability to add SASI indexes to an instance which wasn't configured to support them.  Unsurprisingly that logic was also hiding _other_ schema failures (who could've seen that coming :facepalm: ) so it was yanked and the SASI index generation logic made conditional.

python/adelphi/tests/util/schema_util.py
```diff
@@ -46,22 +50,22 @@ def get_table(keyspace, name):
...
 	indexes = {}
 	# add a regular index
 	regular_index = index(keyspace, name, "regular_index_" + name, None, {"target": "my_column_0"})
 	indexes[regular_index.name] = regular_index
 	# add a CUSTOM index (this must be removed by the anonymizer)
-	custom_index = index(keyspace, name, "custom_index_" + name, "CUSTOM", {"target": "my_column_5", "class_name": "org.apache.cassandra.index.sasi.SASIIndex"})
-	indexes[custom_index.name] = custom_index
+	if sasi:
+		custom_index = index(keyspace, name, "custom_index_" + name, "CUSTOM", {"target": "my_column_5", "class_name": "org.apache.cassandra.index.sasi.SASIIndex"})
+		indexes[custom_index.name] = custom_index
```
absurdfarce, line 68: As mentioned above we must now support schema generation _without_ statements to create SASI indexes since we know they'll fail for at least some Docker images due to a lack of relevant C* config props.

absurdfarce:
Because this PR is blocking other development we're planning on merging this soon in a somewhat imperfect state.  https://github.com/datastax/adelphi/issues/113 has been filed to document some enhancements/fixes to this framework which can be pursued in the future.

grighetto (approved):
Pulled the last commit and confirmed the `cdc` issue is now fixed.

+1 on merging the PR, we still need to work on the results reporting though, as it is it hasn't improved much on the granular feedback as originally intended, but we'll split that into a different task

PR #109, opened 2021-03-04
106 Export to output directory not properly constrained by keyspace

absurdfarce:


absurdfarce:
Still a work-in-progress at this point

absurdfarce:
Ran through a set of adelphi invocations after this change to confirm that we didn't break anything... everything looks solid:

> adelphi export-cql
adelphi --no-anonymize export-cql
adelphi --output-dir=export-cql export-cql
adelphi --output-dir=export-cql --no-anonymize export-cql
adelphi export-cql --no-metadata
adelphi --no-anonymize export-cql --no-metadata
adelphi --output-dir=export-cql export-cql --no-metadata
adelphi --output-dir=export-cql --no-anonymize export-cql --no-metadata
adelphi --keyspaces=testkeyspace,system_views export-cql
adelphi --keyspaces=testkeyspace,system_views --no-anonymize export-cql
adelphi --keyspaces=testkeyspace,system_views --output-dir=export-cql export-cql
adelphi --keyspaces=testkeyspace,system_views --output-dir=export-cql --no-anonymize export-cql
adelphi export-gemini
adelphi --no-anonymize export-gemini
adelphi --output-dir=export-gemini export-gemini
adelphi --output-dir=export-gemini --no-anonymize export-gemini
adelphi --keyspaces=testkeyspace,system_views export-gemini
adelphi --keyspaces=testkeyspace,system_views --no-anonymize export-gemini
adelphi --keyspaces=testkeyspace,system_views --output-dir=export-gemini export-gemini
adelphi --keyspaces=testkeyspace,system_views --output-dir=export-gemini --no-anonymize export-gemini
adelphi export-nb
adelphi --no-anonymize export-nb
adelphi --output-dir=export-nb export-nb
adelphi --output-dir=export-nb --no-anonymize export-nb
adelphi --keyspaces=testkeyspace,system_views export-nb
adelphi --keyspaces=testkeyspace,system_views --no-anonymize export-nb
adelphi --keyspaces=testkeyspace,system_views --output-dir=export-nb export-nb
adelphi --keyspaces=testkeyspace,system_views --output-dir=export-nb --no-anonymize export-nb

jdonenine:
@absurdfarce are all of those various calls of the script that you mention you tested covered by any automated testing of the script at this point or did you have to run all of those manually?

absurdfarce:
@jdonenine This is exactly what prompted my comment in Slack the other day re: automating a baseline integration test.  For now they're all being run manually but it's fairly tedious, in no small measure because the expected behaviour for each of those cases varies quite a bit.  I'd _much_ rather have an automated test which could confirm everything for me.

I was planning on creating a ticket for the baseline testing work anyway... honestly I thought I already had but I cant seem to find it anywhere.  I'll include that list of commands (as well as the expected results) when I do so.  There's prolly some work to do there re: how to mimic running the "adelphi" binary with various flags without actually executing it in a subshell or anything... but that seems like a solvable problem.

absurdfarce:
@jdonenine I filed https://github.com/datastax/adelphi/issues/110 to address the idea of running a baseline integration test (including these invocations) automatically.

absurdfarce:
Note that this isn't completely done yet: we still need to get to a good answer for https://github.com/datastax/adelphi/pull/104/files#diff-1c07f00ef0ebddaf2ea87f34f4191dafc4e113896bc54e48453e3108130a7420R168-R169.  That impl really depends on https://github.com/datastax/adelphi/pull/104 being into master, however, so hold on finishing this up (and by extension merging it) until the other PR is in.

grighetto:
@absurdfarce I'm seeing the following assertion failure on this branch (on version 2.1.22):

*PY 2*
```
AssertionError: {'bindings': {'key1_seq': 'Mod(1000000000); ToString() -> String', 'key2_seq': ' [truncated]... != {'bindings': {'key1_seq': 'Mod(1000000000); ToString() -> String', 'key2_seq': ' [truncated]...
Diff is 2547 characters long. Set self.maxDiff to None to see it.
```

*PY 3*
```
AssertionError: {'bin[473 chars]key1_seq},{key2_dist},{value_dist})', 'tags': [820 chars]o']}} != {'bin[473 chars]key1_dist},{key2_dist},{value_dist})', 'tags':[822 chars]o']}}
Diff is 2638 characters long. Set self.maxDiff to None to see it.
```

grighetto:
Also, it seems none of the tests now actually constrain the export to a single keyspace. It's probably worth it to have an automated test for that scenario too, what do you think? @absurdfarce

absurdfarce:
That's hilarious, really:

```
(tox) [mersault@linux adelphi]$ diff /tmp/tmpl6iCu6/2.1.22-stdout.out tests/integration/resources/nb-schemas/2.1.22.yaml 
13c13
<       values ({key1_dist},{key2_dist},{value_dist})
---
>       values ({key1_seq},{key2_dist},{value_dist})
34c34
<       ({key1_dist},{key2_dist},{value_dist})
---
>       ({key1_seq},{key2_dist},{value_dist})
```

```
(tox) [mersault@linux adelphi]$ more /tmp/tmpl6iCu6/outputDir/2.1.22/testkeyspace/schema
...
blocks:
- name: rampup
  params:
    cl: LOCAL_QUORUM
  statements:
  - rampup-insert: insert into "testkeyspace"."testtable" ("key1","key2","value")
      values ({key1_dist},{key2_dist},{value_dist})
    tags:
      name: rampup-insert
  tags:
    phase: rampup
```

For some reason we're still pulling from the dist there even for primary keys which sure seems... wrong.

I'll get this fixed shortly.

absurdfarce:
Test fixed; it now passes for all C* versions using both py2 and py3.  Apparently I didn't actually run those against all versions after the integration test stuff was merged in... I sure thought I had but there's no way this wouldn't have been caught if that were the case.  A thousand pardons to all involved. :)

python/adelphi/adelphi/nb.py
```diff
@@ -47,11 +47,14 @@ def build_select_statements(keyspace, table):
 
 
 def build_insert_statements(keyspace, table):
+    # Note that both the sequence of column names and column bindings are built off of
+    # the same base sequence (cols below) in order to make sure column names and binding
+    # names line up in the generated CQL.  Order is pretty important here.
     cols = table.columns.values()
-    primary_keys = set(table.primary_key)
+    primary_keys = set([c.name for c in table.primary_key])
```
absurdfarce, line 54: The root cause of the problem with the integration tests.

We built up a set of ColumnMetadata instances returned from the driver and then traverse the list of known ColumnMetadata instances for a table, checking each to see if it's in that set and generating different nosqlbench seq/dist names accordingly.  But for some reason an input ColumnMetadata instance wasn't being identified as a member of the set even though it very definitely was in the set (something I validated two or three different ways).  It works fine to use the column _name_ rather than the column instance directly so I just wound up doing that but... I mean... :huh:

python/adelphi/tests/integration/resources/nb-schemas/2.1.22.yaml
```diff
@@ -10,7 +10,7 @@ blocks:
     cl: LOCAL_QUORUM
   statements:
   - rampup-insert: insert into "testkeyspace"."testtable" ("key1","key2","value")
-      values ({key1_seq},{key2_dist},{value_dist})
+      values ({key1_seq},{key2_seq},{value_dist})
```
absurdfarce, line 13: An interesting fallout from the integration test fixes above.  The use of a compound primary key in our test schema means that we identify both key1 and key2 as in the primary key and therefore both should use the seq rather than the dist.  Technically this is overkill; obviously a seq from a single column would guarantee uniqueness (meaning the original schema is certainly correct).  But this version is not wrong (it's just excessively unique) and it does keep the logic in the impl very straightforward (if primary key use seq, otherwise use dist)... so I updated the reference schemas instead.

absurdfarce:
@grighetto the last few commits add integration tests for the constrained keyspace case (i.e. only a subset of possible keyspaces selected by the user using a CLI option).  This entailed a bit of re-work of our existing integration tests since they were somewhat monolithic and didn't generalize in this case nicely... but that work was coming anyway so it's not a huge problem.

At some point we'll have to address the testing situation more thoroughly; at the moment we loop through all versions of C* _for each test_, which won't scale _at all_ as we add additional integration tests.  Ideally some of these tests could perhaps become unit tests (fleshing out unit test support would also be a good thing) but that's not a real solution.  The integration test framework really should execute all tests for a given C* version and then move on to the next... I'll be filing a ticket for that later (or updating one of the ones we already have if there's an applicable one I'm not thinking of).

Anyways, this problem existed before any work was done on this PR... and it's only modestly exacerbated by the changes here.  I think the tests laid out in this PR will do what we need for now, but I'm gonna start pushing for cleaning up the testing infrastructure more strongly as we actually try to use the thing. :)

grighetto (approved):
Looks good to me 👍
All tests pass now and great addition on the export keyspace tests.
The integration tests take significantly longer now because of all the startup/shutdown dance, but we can improve that in the future.

absurdfarce:
Note: https://github.com/datastax/adelphi/issues/115 has been filed to address the necessary changes in test infrastructure necessary by making these integration tests more granular.

Issue #100, opened 2021-02-04
Add support for additional column types when generating nosqlbench configs from a schema


absurdfarce:
# Feature Request

### Description

https://github.com/datastax/adelphi/issues/86 implemented a base framework for generating nosqlbench configs from an input schema.  To ease the implementation we only added support for "text" types; this should be expanded to support as many CQL types as we reasonably can.

Note that this requires defining nosqlbench distributions for values for each supported type, so there may be a fair amount of twiddling there in order to get things like inetaddress working.

### Describe the solution you'd like

As of this writing the impl has a fairly straightforward method for defining nosqlbench distributions/sequences in code (see https://github.com/datastax/adelphi/pull/96/files#diff-b50d6211d55bd7434d67c456c88581087f9c2c61e279c59891fa474667fc3485R8-R11).  Ideally any impl here would just piggy-back off of that work and expand it somewhat.

Note that the list of types we can expect to see here will have to be extracted from the Python Cassandra driver

absurdfarce:
Note that https://github.com/datastax/adelphi/issues/107 is adding support for a large collection of data types directly, enough to cover commonly used types in schemas.  We still need to cover some additional cases:

* A few types still aren't covered there... consider inetaddress
* Support for the collection types (list, map, set)
* Probably need a good story around the handling (or lack thereof) of UDTs

absurdfarce:
Per comments on https://github.com/datastax/adelphi/issues/107: the PR for that ticket will add support for all basic CQL types.  As of this writing we have everything except "duration" and I expect that to be added within the next day or so.  With that in mind this ticket is intended to add support for the following:

* Counters
* Frozen types
* Collection types (lists, sets and maps)
* Tuples

PR #114, opened 2021-03-23
107 Stepping stone to full type support

absurdfarce:


python/adelphi/bin/adelphi
```diff
@@ -229,11 +229,15 @@ def contribute(ctx, token):
 
 
 @export.command()
+@click.option('--rampup-cycles', type=int, default=1000, help='Number of cycles to use for the nosqlbench rampup phase')
+@click.option('--main-cycles', type=int, default=1000, help='Number of cycles to use for the nosqlbench main phase')
```
absurdfarce, line 233: Added explicit CLI options to specify rampup and main cycles

python/adelphi/adelphi/nb.py
```diff
@@ -47,31 +60,36 @@ def build_select_statements(keyspace, table):
...
     pass
 
 
 class NbExporter(BaseExporter):
 
     def __init__(self, cluster, props):
 
+        self.rampup_cycles = props["rampup-cycles"]
+        self.main_cycles = props["main-cycles"]
+        self.numeric_max = min((self.rampup_cycles + self.main_cycles) * 1000, MAX_NUMERIC_VAL)
```
absurdfarce, line 84: Seemed reasonable to impose a global absolute max on the size of a number within our flow here.  The logic here is the minimum of either our absolute max or a range of 1000 values for each specified cycle (for the distribution case, which requires a wider range of values than the sequence case)

python/adelphi/bin/adelphi
grighetto: NoSQLBench supports template parameters such that you can override these values when launching the workload, which I think is preferred over adding more command-line args to Adelphi itself.

python/adelphi/adelphi/nb.py
```diff
@@ -113,15 +150,28 @@ def __build_main_write_statement(self):
...
+
+
     def __build_schema(self):
         """Really more of a config than a schema, but we'll allow it"""
         root = {}
 
-        rampup_scenario = "run driver=cql tags==phase:rampup cycles===TEMPLATE(rampup-cycles,1000) threads=auto"
-        main_scenario = "run driver=cql tags==phase:main cycles===TEMPLATE(main-cycles,1000) threads=auto"
-        root["scenarios"] = {"default":[rampup_scenario, main_scenario]}
+        root["scenarios"] = {"default":[self.__get_rampup_scenario(), self.__get_main_scenario()]}
```
grighetto, line 172: The scenario name should be templated too, we use different scenario names to distinguish between the workload executions in Grafana (despite using the same workload YAML). We set the names at runtime as a command-line arg.
Take a look at values.yaml, the current sample schema has a template tag around the scenario name.

grighetto (changes requested):

python/adelphi/adelphi/nb.py
absurdfarce: Ah, you're quite right, I did miss that initially.  Will fix shortly.

python/adelphi/bin/adelphi
absurdfarce: Understood, but I don't see why this would be preferred over what's being done here.  Keep in mind that the goal is to re-use the same pattern we used with Gemini i.e. use the adelphi tool to create the config for another app and then call the app with that newly-generated config.  Under that scenario there is no logical or functional difference between specifying the args to nosqlbench at runtime vs. specifying the args to a pre-processor at runtime.

The approach I've taken here also has the (admittedly smaller) benefit of keeping the nosqlbench config simple making it easier to determine exactly what was run, something we couldn't do with a config that was significantly driven by runtime params.

python/adelphi/adelphi/nb.py
absurdfarce: This should be addressed now.

python/adelphi/bin/adelphi
grighetto: Particularly, I think adding options increases complexity, I usually like to ponder if something is really useful before adding just one more param. It's also one more thing we have to test (I don't think that has been covered in this PR).
Since we already have a way to set the ramp-up cycles through nb itself, I don't see the reason why we need to duplicate it on our end.

> Under that scenario there is no logical or functional difference between specifying the args to nosqlbench at runtime vs. specifying the args to a pre-processor at runtime.

That's exactly why I don't think we need the duplication.

>  to determine exactly what was run

That will be kept in helm's values.yaml, which can be exported with the results.

absurdfarce: If we agree that there's no logical difference between the two approaches then it seems to be entirely a cost/benefit trade-off.  And I guess my fundamental argument on that point is that the generated YAML is a _complete_ and  _self-contained_ description of the nosqlbench runs that were executed.  I don't have to consult other files (which hopefully were preserved) or worry about whether environmental information can be reconstructed; the file literally has everything I need.  I can take it and run it immediately in my local nosqlbench instance.  I can preserve it as a build artifact in Jenkins (or something similar) and know I have a way to re-create my nosqlbench run forevermore.  I'd argue it's difficult to overstate the value of such a feature for debugging/auditing/reproducability arguments.

The exchange I have to make to get this feature is the addition of two (maybe three) CLI arguments.  While this does represent some increase in complexity I'd argue that it is much closer to a small increase than a medium one... and nowhere near a large increase.  I'd also observe that your suggestion that a user could reconstruct the info necessary to re-create a nosqlbench run via the nosqlbench runtime args + values.yaml _also_ represents an increase in complexity... one might even argue a larger increase than what's done here.

Finally, I'm not sure I'd refer to this as a duplication of functionality in nosqlbench.  Our YAML generator takes some number of args and uses the value of those args to generate some of the YAML... that's not redundant to the nosqlbench support for parameterizing values, nor would I want it to be.

In the end I just don't see a clear benefit to leaving in the parameters for nosqlbench... it just doesn't seem to buy us nearly enough.

python/adelphi/adelphi/nb.py
absurdfarce: I thought about this some over the weekend and I'm not sure this is the right solution in the end.  It seems like we should make this consistent with the cycles args discussed above; we should either support parameterization or not but I'm not sure there's a clear case for some things being specified via "adelphi" args and some via nosqlbench args.  I'm gonna take a run at converting this to following the pattern used for the cycles elsewhere.

absurdfarce: After pondering this a bit more I'm going to revert this work and go back to the idea of a scenario name as a nosqlbench param.  The impl in place now isn't really right; we'd need to run the generation process twice, once with a scenario name of "source" and once with a name of "target", simply to generate two configs which are otherwise identical.  This seems like overkill for a label which doesn't impact test functioning in any meaningful way.  We do need to provide a default value so that _some_ label is available for nosqlbench runs but as long as that's in place we should be good.

absurdfarce:
After https://github.com/datastax/adelphi/pull/114/commits/4b236a48195b41ae6928454b34169b84f72dfb17 we now support everything except for counters, frozen fields and collection types (lists, sets and maps).  These are the items which will be handled in https://github.com/datastax/adelphi/issues/100; I'll update the description of that ticket accordingly.

The only exception to the above is the duration type added in v5 of the CQL protocol.  nosqlbench doesn't have support for generating this type via virtdata at the moment but Shooky is (literally) on the case as I write this... expectation is that new versions of nosqlbench available as of tomorrow will include such support.  Once it's added in I'll add it to this PR as well.

python/adelphi/adelphi/nb.py
grighetto: > The impl in place now isn't really right; we'd need to run the generation process twice, once with a scenario name of "source" and once with a name of "target", simply to generate two configs which are otherwise identical. 

That goes back to my original thought that we should specify whatever parameters through nosqlbench such that we don't have to regenerate the whole file. We just pick a default and override it at runtime if necessary.
Doing the same for the cycle args would be at least consistent.

python/adelphi/bin/adelphi
grighetto: I was referring to usability complexity. This will be available in the adelphi lib and people may use it standalone. Figuring out and setting a bunch of parameters in Adelphi increases friction.
On our side, it's an extra param to test and inconsistent with other params we already have to set through nosqlbench (scenario name).

I'll open a separate ticket so we can add the cycles arg to the nb template and unblock this PR.

grighetto (approved):
This PR removed support for setting cycles through NoSQLBench, we should add that back in such that we don't have to regenerate the YAML. It will also simplify our command-line tool.
Otherwise, PR looks good, approving to unblock other work 👍

python/adelphi/adelphi/nb.py
```diff
@@ -6,10 +6,30 @@
...
+CQL_TYPES["boolean"] = "Mod({}); ToBoolean() -> java.lang.Boolean"
+CQL_TYPES["varchar"] = "Mod({}); ToString() -> String"
+CQL_TYPES["varint"] = "ModuloToBigInt({}) -> java.math.BigInteger"
+CQL_TYPES["blob"] = "Mod({}); ToByteBuffer() -> java.nio.ByteBuffer"
+CQL_TYPES["timestamp"] = "Mod({}); ToDate() -> java.util.Date"
+CQL_TYPES["date"] = "Mod({}); LongToLocalDateDays() -> com.datastax.driver.core.LocalDate"
+CQL_TYPES["time"] = "Mod({}); ToLong() -> long"
+CQL_TYPES["uuid"] = "Mod({}); ToHashedUUID() -> java.util.UUID"
+CQL_TYPES["timeuuid"] = "Mod({}); ToTimeUUIDMax() -> java.util.UUID"
+CQL_TYPES["inet"] = "Mod({}); ToInetAddress() -> java.net.InetAddress"
```
absurdfarce, line 32: This has been discussed elsewhere but it seemed worth adding a note here as well since this could easily be a place somebody might land while doing some investigation.

Note that we are deliberately not including support for the "duration" type introduced in Cassandra 4.0.  Support for generating sequences of this type wasn't built into nosqlbench at the time we started work.  The functionality has since been added (see https://github.com/nosqlbench/nosqlbench/releases/tag/nosqlbench-4.15.33) but since it wasn't ubiquitous at this writing we were left with a few options:

* Include support for the type and specify that users who wish to run Adelphi against schemas with duration types for primary keys must use nosqlbench 4.15.33 or greater
* Leave the type unsupported for now (with the understanding it will be added in later)

The second option was judged to be less problematic than the first, although it's certainly possible this could be revisited as we gather additional data.

Issue #116, opened 2021-03-25
setup.py should be a bit more flexible with it's versions


absurdfarce:
# Bug Report
### Description
Originally reported by @aholmberg.

Current version of Adephi on pypi (https://pypi.org/project/adelphi/0.1.0/) is a bit too specific w.r.t. it's dependency versions.  This from https://github.com/datastax/adelphi/blob/d3aab446a8cd2181c7c5734e2a7075dfbe21cbf0/python/adelphi/setup.py#L7-L11):

```
dependencies = [
    'cassandra-driver ~= 3.24.0',
    'click ~= 7.1.0',
    'PyGithub ~= 1.45'
    ]
```

Adam pointed out that 3.24 is no longer the most recent version of cassandra-driver.  There's no real reason to constrain these dependencies to this degree... we should make better use of compatible versions.

┆Issue is synchronized with this [Jira Task](https://datastax-oss.atlassian.net/browse/AD-33) by [Unito](https://www.unito.io)
┆Issue Number: AD-33

absurdfarce:
Worth noting: cassandra-driver 3.24.0 won't connect to current C* trunk so this definitely needs to happen before release.  Users of Adelphi + C* trunk in the wild may also see a problem here but any such user can always update cassandra-driver manually.

Current thinking is that there's no need for a 0.1.1 release to address this issue since the affected population is likely very small and a functioning workaround exists.

absurdfarce:
Tested Python2/3 using tox.  Note that if you do so you'll want to delete the existing venvs in the .tox directory; we're changing dependencies above and those will only affect what happens when the venv is created.  If you don't rebuild the venv you'll just re-use what you have, meaning you won't test with the new cassandra-driver version.

PR #117, opened 2021-03-29
116 setup.py should be a bit more flexible with it's versions

absurdfarce:


absurdfarce:
Please note the [comment](https://github.com/datastax/adelphi/issues/116#issuecomment-809667726) on the ticket itself re: tox venv's when testing

python/adelphi/setup.py
```diff
@@ -5,10 +5,10 @@
 PY3 = sys.version_info[0] == 3
 
 dependencies = [
-    'cassandra-driver ~= 3.24.0',
-    'click ~= 7.1.0',
+    'cassandra-driver ~= 3.24',
```
grighetto, line 8: I know this has come up multiple times before, but I'm a bit torn between being flexible with the driver version versus ensuring non-anonymized info won't leak.
What if a newer driver version adds more elements to the CQL output that we're not anonymizing? Or if it changes how it handles the object references internally?

I think the ideal approach would be to detach from the `cluster.metadata` object that we rely on straight out of the driver and instead deep clone `metadata` and copy over only the attributes we're anonymizing. That way we wouldn't have to worry about driver changes.

absurdfarce: I don't think the management of the cluster.metadata object is the problem in the scenario you describe @grighetto.  The fns in adelphi.anonymize swap names around in the various metadata objects returned by the driver.  Whether that's the actual cluster.metadata object itself or a clone of it doesn't really change much... you're still making some assumptions about the internal structures of the metadata object.  Well, I mean you kinda are... you're assuming certain fields exist and that they're writable, and while I agree that that does make you dependent upon the driver API I don't think you make the situation better by using a clone instead.

It seems like the way to accomplish what you're after would be to create some new structure (let's say just a map), populate it from metadata information and pass that along to anonymization.  If I read your comment right this is closer to what you were suggesting.  You could then insulate anonymization against changes in the driver API... but what good does that really do for you?  You've just pushed the problem up a step; now changes to the driver API break your extraction process but they don't break anonymization.  Is that... better?

To make matters worse all of this is overshadowed by the fact that we generate the CQL by calling the method export_as_string() on metadata objects.  Now we're no longer talking about the structure of metadata objects; we're now invoking logic.  That logic could do anything and could change between releases.  The changes described above provide exactly zero protection against that kind of change.  Neither, unfortunately, do our existing unit tests around anonymization.

So why do I not share your concern about these changes?  Two reasons really.

First, we also have integration tests around CQL generation which function as end-to-end tests.  If new material were added in newer driver versions any run against those new versions would indicate a mismatch.

Second, there is no active feature development on the Python driver (or any driver for that matter) at the moment because no team is working on them.  So it's extremely unlikely that any such feature would ever get scheduled let alone released.

So it seems like the concern here requires (a) work from a team that doesn't exist [1] to publish a new version that (b) breaks anonymization in some way and (c) somehow isn't subjected to our integration tests.  While I admit that such a scenario is logically possible the actual probability of all of that happening seems very, very small to me.

[1] This is a bit of an overstatement; a few minor patches are making it in and I think there's still one contractor working on some things.  But it's not much of an overstatement.

grighetto: > Whether that's the actual cluster.metadata object itself or a clone of it doesn't really change much

The "clone" would only contain the attributes we anonymize. If a newer driver version adds some information that we're not anonymizing, it wouldn't appear in the resulting CQL output.

> changes to the driver API break your extraction process but they don't break anonymization. Is that... better?

Definitely, the worst thing that could possibly happen is leaking information.

> That logic could do anything and could change between releases. The changes described above provide exactly zero protection against that kind of change.

As long as the clone contain only the information we want and we're not connected to the cluster, we should be safe with regards to leaking.

> we also have integration tests around CQL generation which function as end-to-end tests

How does that help adelphi installations out there that will pick up a newer driver version automatically?

> there is no active feature development on the Python driver

I wouldn't make any kind of technical decision based on that.

> While I admit that such a scenario is logically possible the actual probability of all of that happening seems very, very small to me.

Again, we must prevent any possibility of leaking info.

absurdfarce:
Per offline conversation with @grighetto the intent is to move forward with this ticket and consider the question of compatible versioning without the pressure of needing to get something done for a release.  To that end I've filed https://github.com/datastax/adelphi/issues/121 to continue the discussion.  @grighetto would you mind giving me an approval on this PR so I can get this merged?

grighetto (approved):
Approving this PR since we need it for C* 4.0 compatibility. 👍 
Enhancement work to shield us from driver changes will come up shortly after.

PR #122, opened 2021-04-14
Some minor test improvements

absurdfarce:


python/adelphi/tests/integration/__init__.py
```diff
@@ -87,11 +87,16 @@ def createSchema(self, session=None):
...
+    def getContainer(self, client, version):
+        return client.containers.run(name="adelphi", remove=True, detach=True, ports={9042:9042}, image="cassandra:{}".format(version))
+
+
     def runTestForVersion(self, version=None):
         log.info("Testing Cassandra version {}".format(version))
 
         client = docker.from_env()
-        container = client.containers.run(name="adelphi", remove=True, detach=True, ports={9042:9042}, image="cassandra:{}".format(version))
+        container = self.getContainer(client, version)
```
absurdfarce, line 99: When running the integration tests I'll often see random errors of this variety:

```
E       docker.errors.APIError: 409 Client Error for http+docker://localhost/v1.39/containers/create?name=adelphi: Conflict ("Conflict. The container name "/adelphi" is already i
n use by container "6fdce64f40d9b48a554b005a17550adaf7082d06db7aa0fa69f6064354eafa50". You have to remove (or rename) that container to be able to reuse that name.") 
```

This appears to be caused by some kind of underlying race within the Docker code (or it's interaction with the testing regime) which tries to create a new test before the prior one has been removed.  My hope was that retry logic would do something to minimize these problems and that looks to be at least somewhat true based on a few runs since the change.

There's more work to be done here; this isn't intended to be some kind of comprehensive solution.

absurdfarce: The obvious way to avoid this problem is to give each image a distinct name based on C* version and/or test to run.  I was concerned about this possibly leading to similar errors (say port availability issues) if images failed to shut down and other images tried to re-use locked up resources.  That might not be as big a concern as I'd originally feared... something to consider anyway.

grighetto (approved):
LGTM 👍

Issue #79, opened 2020-12-17
Build C* and the Management API in the same pod


grighetto:
For testing https://github.com/datastax/management-api-for-apache-cassandra/issues/58 in Adelphi, it became required to build C* in the same pod that builds the Management API, such that the jars generated with `ant mvn-install` are available to the API builder.

And since we'll build it all in the same pod, we'll be able to generate the API image straight away, without the intermediate C* image, which simplifies the workflow a little bit too.

┆Issue is synchronized with this [Jira Task](https://datastax-oss.atlassian.net/browse/AD-58) by [Unito](https://www.unito.io)
┆Issue Number: AD-58

PR #125, opened 2021-04-27
#79 CRD Upgrade for C* 4.0 compatibility

grighetto:


helm/adelphi/templates/cassandra-mgmt-dockerfile-configmap.yaml
```diff
@@ -5,222 +5,81 @@ metadata:
...
-      MGMT_API_JAR="$(find "/opt/mgmtapi" -name *server*.jar)"
-
-      echo "Running" java ${MGMT_API_JAVA_OPTS} -Xms128m -Xmx128m -jar "$MGMT_API_JAR" $MGMT_API_ARGS
-      exec /tini -g -- gosu cassandra java ${MGMT_API_JAVA_OPTS} -Xms128m -Xmx128m -jar "$MGMT_API_JAR" $MGMT_API_ARGS
-
-    fi
-
-    exec "$@"
\ No newline at end of file
+{{ .Files.Get "config/mgmt-api/docker-entrypoint.sh" | indent 4 }}
```
absurdfarce, line 85: Should we consider doing something like this with the Dockerfile as well?  I love the idea of getting this out of inline YAML.

absurdfarce (approved):
Observed clean run with current k3d install.  Validated 3.11.7 and 4.0-rc1 as source and target versions (respectively) so we're good with 4.0-rc1 now (as expected).  Observed expected results from results-server at conclusion of run.

Issue #119, opened 2021-04-06
Integrity testing results via Grafana dashboards


jdonenine:
# Feature Request
### Description
As part of the overall goal of the broader #111  one aspect of visualizing results is via Grafana dashboards.  It would be great to see a few different things made available at the end of the testing workflow:

* [ ] Pre-built dashboards tailored to visualizing the integrity testing results generated by Adelphi
* [ ] Grafana made available within the cluster running at the completion of the overall workflow
* [ ] Instructions/scripts for exposing access to the grafana service running within the cluster

### Additional context

This issue was spun out of #111

┆Issue is synchronized with this [Jira Task](https://datastax-oss.atlassian.net/browse/AD-31) by [Unito](https://www.unito.io)
┆Issue Number: AD-31

Issue #91, opened 2021-02-01
Build comparison dashboard for Adelphi


grighetto:
# Feature Request

### Description

We're adding a Grafana Dashboard to the Adelphi stack such that we can visualize performance data in realtime from both clusters (source and target). But the default dashboard that comes with NoSQLBench wasn't really designed for that kind of comparison need, so we can do a better job by creating a custom board that shows the charts side-by-side (and maybe an alternative overlay view) with sensible metrics that makes sense for Adelphi.

### Describe the solution you'd like

- Understand and select the metrics that we want to display by default;
- Build a new dashboard in Grafana, export it to JSON and import it into Adelphi.

PR #126, opened 2021-04-27
#119 Display Gemini integrity test results in Grafana

grighetto:
BRANCHED OFF #125 - CLEAN DIFF AVAILABLE AT https://github.com/datastax/adelphi/compare/79-enable-exec-against-latest-cass...119-gemini-results-in-grafana _(just until parent branch is merged)_


helm/adelphi/templates/nosqlbench-jobs.yaml
```diff
@@ -29,21 +29,33 @@ spec:
...
+        mountPath: /workspace
       - name: results-pv
         mountPath: /results
+    initContainers:
+    - name: extract-schema
+      image: python:3.10.0b1-alpine3.13
+      command: ["sh", "-c", "-e"]
+      args:
+        - pip install adelphi;
+          adelphi --hosts="{{ .Values.source.clusterName }}-{{ .Values.source.dc }}-service" --username="{{ .Values.source.clusterName }}-superuser" --password="`cat /secret/password`" --keyspaces="{{ .Values.gemini_schema_keyspace }}" --no-anonymize export-nb > /workspace/nosqlbench_source.yaml;
```
absurdfarce, line 42: I'm a little worried about bringing this in until we have full support for all the column types (i.e. https://github.com/datastax/adelphi/issues/100).

helm/adelphi/templates/gemini-results.yaml
```diff
@@ -0,0 +1,36 @@
...
+            shift;
+            cat /results/results.json |
+            jq -r ".result.$METRIC" |
+            xargs -i echo "gemini.$METRIC {} $TIMESTAMP" |
+            nc graphite-service 9109;
+          };
+          push "write_ops";
+          push "write_errors";
+          push "read_ops";
+          push "read_errors";
```
absurdfarce, line 29: This feels like the kind of thing that would be better expressed via a script that could just be run somewhere rather than an inline collection of shell commands like this

absurdfarce:
Worth noting that a new field (Data Integrity) was added here but we didn't remove any.  And since the old panel was setup to display a number of graphs which fit exactly into a couple rows we now have a single row at the bottom with just one graph (service time range) in it:

grafana-snapshot

Perhaps we should just... yank this?

absurdfarce:
Perhaps not something we need to sort out right now but I'm a _little_ worried that it may not be immediately obvious what the contents of the "Data Integrity" panel are referring to.  The terms used there (write/read ops and/or errors) are all pretty generic... they don't clearly indicate that Adelphi checked the results of it's source/target compares and found all was good.

helm/adelphi/templates/nosqlbench-jobs.yaml
grighetto: @absurdfarce What is in master right now doesn't even use the provided schema, so this is obviously better. We have other tickets to handle failures/logging scenarios.

If there's still a concern, let's just make our docs more clear, but let's not hold the feature back.

helm/adelphi/templates/gemini-results.yaml
grighetto: @absurdfarce I like shell commands in workflow templates, it makes it pretty easy to track where/when it gets executed. Between that and a separate script, I don't see much difference.
The ideal solution would be patching Gemini to send the metrics directly as it generates them, so we'd actually have a time-series instead of just a single data point at the end. I'm planning to tackle that along with other Gemini work we have planned.
In the meantime, this should be fine.

grighetto:
> And since the old panel was setup to display a number of graphs which fit exactly into a couple rows we now have a single row at the bottom with just one graph Perhaps we should just... yank this?

@absurdfarce Are you suggesting we should remove a panel because the layout doesn't look good on screen? We have planned work to improve the Grafana dashboard, we can probably handle layout issues there, but I'd say we should display whatever metric we think is relevant (that's a separate discussion).

grighetto:
> they don't clearly indicate that Adelphi checked the results of it's source/target compares and found all was good

We can probably add a top-row guide similar to what NoSQLBench had originally to explain each panel. But that should probably happen after the new Grafana dashboard is done.

helm/adelphi/templates/gemini-results.yaml
jdonenine: I was debating something similar with myself 😄  recently w.r.t. to GH Actions, ultimately I came to preferring the commands inline in the action, similar to what you've got here.  What makes the decision easier I think is would the contents of that script be used by anything but this?  If it would be, then a separate script makes sense, if not, I like the ease of being able to understand what will happen directly from the template.

absurdfarce:
> @absurdfarce Are you suggesting we should remove a panel because the layout doesn't look good on screen? We have planned work to improve the Grafana dashboard, we can probably handle layout issues there, but I'd say we should display whatever metric we think is relevant (that's a separate discussion).

I was suggesting we should do something to fix the presentation.  Yanking the last graph is one way to solve this.  So is moving to a 3x3 instead of a 2x4 + 1 grid.  And there are probably other ways to solve it as well.

grighetto:
> I was suggesting we should do something to fix the presentation. Yanking the last graph is one way to solve this. So is moving to a 3x3 instead of a 2x4 + 1 grid. And there are probably other ways to solve it as well.

@absurdfarce thanks for clarifying, but let's move that discussion to https://github.com/datastax/adelphi/issues/91

absurdfarce (approved):

PR #127, opened 2021-04-28
101 Integrate nosqlbench config generation into k8s

absurdfarce:
Add Docker image for alpine + installed Adelphi CLI and use this for invocations within the Helm workflow.

Note that this PR should **NOT** be merged until we've pushed the new image to Dockerhub and adjusted the "image" usages below accordingly.


helm/adelphi/templates/nosqlbench-jobs.yaml
```diff
@@ -26,24 +26,36 @@ spec:
...
         mountPath: /secret
         readOnly: true
-      - name: workload
-        mountPath: /config/workload.yaml
-        subPath: workload.yaml
       - name: results-pv
         mountPath: /results
+    initContainers:
+    - name: extract-config
+      image: registry:5000/adelphi:0.2.0
```
absurdfarce, line 38: Original plan was to continue using the "pip install adephi + use adelphi" model seen in the Gemini example above before this change.  Had to shift away from this due to some dependency changes.

While working on this PR PyGithub released v1.55 which introduced a dependency on pynacl for supporting Github secrets.  pynacl in turn relies on native code to interact with libsodium, which meant the base image we used here needed to have gcc + some other dependencies installed.  That in turn led to the cassandra-driver install trying to compile it's uring extension now that it could do so... and suddenly the pip install of "adelphi" was taking ~20m due to all the added dependencies.

Rather than continue on with this path it seemed _much_ simpler to just create an image with "adelphi" already installed.  In addition to the _massive_ time savings this also has the effect of fixing the dependency versions used by "adelphi" _for usages within the workflow_.  These dependencies still float in setup.py (which we want) but we still get clear, fixed behaviour within Helm (which we also want).

docker/adelphi/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM python:3.9-alpine3.12
+
+RUN apk add build-base libffi-dev python3-dev libev-dev
```
absurdfarce, line 3: Need to check on exactly what's in these packages but my suspicion would be we could subsequently remove some/all of them after compilation in order to cut down on image size.

absurdfarce:
Including gcc is _definitely_ the wrong thing to do.  Closing this out in favor of a better way of creating the image.

PR #148, opened 2021-06-02
129 make template vars great again

absurdfarce:


python/adelphi/adelphi/nb.py
```diff
@@ -1,15 +1,17 @@
 import logging
 import yaml
 
-from itertools import chain
-
```
absurdfarce, line 5: Just a bit of cleanup... apparently this wasn't used anymore

python/adelphi/adelphi/nb.py
```diff
@@ -89,7 +91,12 @@ def __init__(self, cluster, props):
 
         self.rampup_cycles = props["rampup-cycles"]
         self.main_cycles = props["main-cycles"]
-        self.numeric_max = min((self.rampup_cycles + self.main_cycles) * 1000, MAX_NUMERIC_VAL)
+        self.scenario_name = props["scenario-name"]
+        self.numeric_max = \
+            min((self.rampup_cycles + self.main_cycles) * 1000, MAX_NUMERIC_VAL) \
+            if self.rampup_cycles and self.main_cycles \
+            else MAX_NUMERIC_VAL
```
absurdfarce, line 98: This reflects the maximum size we can get for numeric values.  If the number of cycles is specified we can use this to constrain the number of values in order to increase the likelihood of hits when we do selects on primary keys.

python/adelphi/adelphi/nb.py
```diff
@@ -172,7 +181,7 @@ def __build_schema(self):
         """Really more of a config than a schema, but we'll allow it"""
         root = {}
 
-        root["scenarios"] = {"TEMPLATE(scenarioname,default)":[self.__get_rampup_scenario(), self.__get_main_scenario()]}
+        root["scenarios"] = {self.scenario_name or TEMPLATEVAR_SCENARIO_NAME:[self.__get_rampup_scenario(), self.__get_main_scenario()]}
```
absurdfarce, line 184: Note that in all three cases the template var is the default case.  If the user doesn't explicitly specify a val via the CLI args we use the template var instead.

python/adelphi/tests/integration/__init__.py
```diff
@@ -22,6 +22,20 @@
...
+    def cqlReferenceSchema(self, version):
+        return "tests/integration/resources/cql-schemas/{}.cql".format(version)
+
+
+    def nbReferenceYaml(self, version):
+        return "tests/integration/resources/nb-schemas/{}.yaml".format(version)
+
+
+    def nbBaseSchema(self):
+        return "tests/integration/resources/nb-base-schema.cql"
```
absurdfarce, line 35: Some file cleanup.  Patterns were developing in the tests around referencing base + expected CQL/YAML but many of the tests were still manipulating raw paths.  Goal here is to consolidate that logic in the framework so that it only has to be implemented once.

python/adelphi/adelphi/nb.py
```diff
@@ -185,4 +194,4 @@ def __build_schema(self):
         main_write_block = {"name":"main-write", "tags":{"phase":"main", "type":"write"}, "params":cl_ratio_map, "statements": [self.__build_main_write_statement()]}
         root["blocks"] = [rampup_block, main_read_block, main_write_block]
 
-        return yaml.dump(root, default_flow_style=False)
+        return yaml.safe_dump(root, default_flow_style=False)
```
absurdfarce, line 197: We were seeing cases in which the YAML dump was writing a serialized Python Unicode string here which wound up with something like "!!python/unicode foo".  PyYAML supports serializing arbitrary Python types; safe_dump() requires the output to be a base string, which is exactly what we want here.

absurdfarce:
The testing bits here are going to be in a bad way in light of https://github.com/datastax/adelphi/pull/152.  I think it's prolly best to get that PR merged and then re-work this guy so that he plays nice with the new integration test regime.

PR #154, opened 2021-06-11
58 CI automation of test suite

absurdfarce:


python/adelphi/adelphi/anonymize.py
```diff
@@ -15,7 +15,6 @@
 
 # Functions and constants related to the anonymization process
 from adelphi.store import get_standard_columns_from_table_metadata
-import re
```
absurdfarce, line 18: Randomly discovered unused import, presumably a holdover from previous changes

python/adelphi/adelphi/gh.py
```diff
@@ -22,7 +22,6 @@
 
 from github import Github
 
-logging.basicConfig(level=logging.INFO)
```
absurdfarce, line 25: This (and store.py below) should definitely _not_ be setting up their own logging... they should be leveraging logs that are configured somewhere else.

python/adelphi/adelphi/store.py
```diff
@@ -39,20 +40,26 @@
...
-    if username is not None and password is not None:
-        auth_provider = PlainTextAuthProvider(username=username, password=password)
-    return auth_provider
+    """Instantiate auth provider if credentials have been provided"""
+    if username is None or password is None:
+        return None
+    return PlainTextAuthProvider(username=username, password=password)
 
 
+@retry
```
absurdfarce, line 49: Annotated to make this a tenacity-aware function but we don't provide global configs here since those are largely context-specific.  Individual callers can provide args appropriate for their specific use; see the example in test-adelphi to see what this looks like.

python/adelphi/bin/test-adelphi
```diff
@@ -74,13 +109,13 @@ def runtests(cassandra, python, pytest):
             # exiting all the things
             try:
                 tox.cmdline(tox_args)
-            except SystemExit:
-                pass
+            except SystemExit as exc:
+                exitCodes.append(exc.code)
```
absurdfarce, line 113: The exit code that would be returned by tox lives in SystemExit.code.  We need to trap this to determine how many of our tox "invocations" resulted in successful tests.

python/adelphi/test-requirements.txt
```diff
@@ -1,5 +1,2 @@
-click ~= 7.1
-cassandra-driver ~= 3.24
 docker ~= 4.4
-tenacity ~= 7.0
```
absurdfarce, line 4: Now that test-adelphi is a full-fledged part of the Python package (see setup.py change above) there isn't any need to manage it's dependencies via test requirements.  This change was one of the significant arguments for making test-adelphi part of the package.

absurdfarce:
Look at "Checks" on this PR to see the output of running a complete test suite on the code in this PR.

.github/workflows/adelphi-python.yaml
```diff
@@ -0,0 +1,29 @@
+name: adelphi Python
+
+on:
+  push:
+    branches: [ 115-breadth-first-testing ]
+  pull_request:
+    branches: [ 115-breadth-first-testing ]
```
absurdfarce, line 7: A temporary config (since this is branched off of the feature branch changing the integration test framework).  This will need to be changed to the default branch before we actually merge it in.

absurdfarce:
Closing in favor of https://github.com/datastax/adelphi/pull/160

PR #156, opened 2021-06-17
Update .gitignore to exclude generated tox file

jdonenine:


.gitignore
```diff
@@ -56,6 +56,7 @@ coverage.xml
 *.py,cover
 .hypothesis/
 .pytest_cache/
+python/adelphi/tox.ini
```
absurdfarce, line 59: I'm wondering if we want to be even more aggressive and not include tox.ini anywhere in the repo... although as a practical matter what you have here will cover the overwhelmingly common case.  I'm fine with this.

PR #159, opened 2021-06-21
113 Improve reporting of test failures for CQL tests, take two

absurdfarce:


python/adelphi/tests/integration/test_cql.py
```diff
@@ -64,19 +51,23 @@ def buildSchema(self):
...
-            logCqlDigest(referencePath, refOnlySet)
-        compareOnlySet = compareSet - referenceSet
+            for digest in referenceOnlySet:
+                log.info(referenceDict[digest])
+        compareOnlySet = compareDigests - referenceDigests
         if len(compareOnlySet) > 0:
             log.info("Statements in compare file {} but not in reference file {}:".format(comparePath, referencePath))
-            logCqlDigest(comparePath, compareOnlySet)
+            for digest in compareOnlySet:
+                log.info(compareDict[digest])
```
absurdfarce, line 68: I'm not really sure the explicitly logging we're doing here is necessary now.

Given the following diff (to introduce a CQL mismatch):

```
diff --git a/python/adelphi/tests/integration/resources/cql-schemas/4.0-rc1.cql b/python/adelphi/tests/integration/resources/cql-schemas/4.0-rc1.cql
index dbe0a97..c875e18 100644
--- a/python/adelphi/tests/integration/resources/cql-schemas/4.0-rc1.cql
+++ b/python/adelphi/tests/integration/resources/cql-schemas/4.0-rc1.cql
@@ -6,7 +6,7 @@ CREATE TYPE IF NOT EXISTS ks_0.udt_0 (
 );
 
 CREATE TYPE IF NOT EXISTS ks_0.udt_1 (
-    fld_2 frozen<udt_2>,
+    fld_22 frozen<udt_2>,
     fld_3 frozen<map<frozen<udt_2>, frozen<udt_2>>>,
     fld_4 frozen<list<frozen<udt_0>>>,
     fld_5 frozen<tuple<frozen<udt_2>, frozen<udt_0>>>
```

the blocks above print the following in the log (as displayed by tox):

```
INFO     adelphi:cassandra_util.py:23 Creating schema on Cassandra cluster from file /tmp/tmpoBGk1E/base-schema.cql
INFO     adelphi:test_cql.py:61 Statements in reference file tests/integration/resources/cql-schemas/4.0-rc1.cql but not in compare file /tmp/tmpoBGk1E/4.0-rc1-stdout.out:
INFO     adelphi:test_cql.py:63 CREATE TYPE IF NOT EXISTS ks_0.udt_1 ( fld_22 frozen<udt_2>, fld_3 frozen<map<frozen<udt_2>, frozen<udt_2>>>, fld_4 frozen<list<frozen<udt_0>>>, f
ld_5 frozen<tuple<frozen<udt_2>, frozen<udt_0>>> );                  
INFO     adelphi:test_cql.py:66 Statements in compare file /tmp/tmpoBGk1E/4.0-rc1-stdout.out but not in reference file tests/integration/resources/cql-schemas/4.0-rc1.cql:
INFO     adelphi:test_cql.py:68 CREATE TYPE IF NOT EXISTS ks_0.udt_1 ( fld_2 frozen<udt_2>, fld_3 frozen<map<frozen<udt_2>, frozen<udt_2>>>, fld_4 frozen<list<frozen<udt_0>>>, fl
d_5 frozen<tuple<frozen<udt_2>, frozen<udt_0>>> ); 
```

Note, however, that we've now switched to using sets of CQL statements for the assertEqual() call below, so the test actually fails with the following message:

```
ests/integration/test_cql.py:70: in compareToReferenceCql                                                                                                                        
    self.assertEqual(set(referenceDict.values()), set(compareDict.values()))                                                                                                      
E   AssertionError: Items in the first set but not the second:                                                                                                                    
E   'CREATE TYPE IF NOT EXISTS ks_0.udt_1 ( fld_22 frozen<udt_2>, fld_3 frozen<map<frozen<udt_2>, frozen<udt_2>>>, fld_4 frozen<list<frozen<udt_0>>>, fld_5 frozen<tuple<frozen<ud
t_2>, frozen<udt_0>>> );'                                                                                                                                                         
E   Items in the second set but not the first:                                                                                                                                    
E   'CREATE TYPE IF NOT EXISTS ks_0.udt_1 ( fld_2 frozen<udt_2>, fld_3 frozen<map<frozen<udt_2>, frozen<udt_2>>>, fld_4 frozen<list<frozen<udt_0>>>, fld_5 frozen<tuple<frozen<udt
_2>, frozen<udt_0>>> );'  
```

It's not clear to me that the explicit logging we do here is anything but repetitive.

grighetto:
While this is a great improvement over what we currently have in the new test framework, it's still not as easy to parse as `diff` was. For example, I changed a single letter in a data type and I got this ouput:

> E   AssertionError: Items in the first set but not the second:
> E   "CREATE TABLE IF NOT EXISTS ks_0.tbl_0 ( col_0 ascii, col_1 bigint, col_2 blob, col_3 boolean, col_4 frozen<udt_1>, col_5 map<int, int>, col_6 set<int>, col_7 frozen<tuple<int, frozen<tuple<text, bouble>>>>, col_8 text, col_9 timestamp, col_10 timeuuid, col_11 uuid, col_12 text, col_13 varint, col_14 decimal, col_15 double, col_16 float, col_17 inet, col_18 int, col_19 list<int>, col_20 frozen<udt_2>, PRIMARY KEY ((col_0, col_1), col_2, col_3) ) WITH CLUSTERING ORDER BY (col_2 ASC, col_3 ASC) AND bloom_filter_fp_chance = 0.01 AND caching = {'keys': 'ALL', 'rows_per_partition': 'NONE'} AND comment = '' AND compaction = {'class': 'org.apache.cassandra.db.compaction.SizeTieredCompactionStrategy', 'max_threshold': '32', 'min_threshold': '4'} AND compression = {'enabled': 'false'} AND crc_check_chance = 1.0 AND dclocal_read_repair_chance = 0.1 AND default_time_to_live = 0 AND gc_grace_seconds = 864000 AND max_index_interval = 2048 AND memtable_flush_period_in_ms = 0 AND min_index_interval = 128 AND read_repair_chance = 0.0 AND speculative_retry = '99PERCENTILE';"
> E   Items in the second set but not the first:
> E   "CREATE TABLE IF NOT EXISTS ks_0.tbl_0 ( col_0 ascii, col_1 bigint, col_2 blob, col_3 boolean, col_4 frozen<udt_1>, col_5 map<int, int>, col_6 set<int>, col_7 frozen<tuple<int, frozen<tuple<text, double>>>>, col_8 text, col_9 timestamp, col_10 timeuuid, col_11 uuid, col_12 text, col_13 varint, col_14 decimal, col_15 double, col_16 float, col_17 inet, col_18 int, col_19 list<int>, col_20 frozen<udt_2>, PRIMARY KEY ((col_0, col_1), col_2, col_3) ) WITH CLUSTERING ORDER BY (col_2 ASC, col_3 ASC) AND bloom_filter_fp_chance = 0.01 AND caching = {'keys': 'ALL', 'rows_per_partition': 'NONE'} AND comment = '' AND compaction = {'class': 'org.apache.cassandra.db.compaction.SizeTieredCompactionStrategy', 'max_threshold': '32', 'min_threshold': '4'} AND compression = {'enabled': 'false'} AND crc_check_chance = 1.0 AND dclocal_read_repair_chance = 0.1 AND default_time_to_live = 0 AND gc_grace_seconds = 864000 AND max_index_interval = 2048 AND memtable_flush_period_in_ms = 0 AND min_index_interval = 128 AND read_repair_chance = 0.0 AND speculative_retry = '99PERCENTILE';"

What do you think about comparing line by line instead? I think that's a minor changed that would help a lot with debugging.

grighetto:
Another issue I'm having is that my tox.ini file doesn't take effect anymore. I used to manually edit it before to set the correct python cli name for my local environment.
I think that was introduced it in #152 and the tox config became hardcoded.

python/adelphi/tests/integration/test_cql.py
absurdfarce: No longer applicable

absurdfarce:
Re: line-by-line diffs... I also wasn't super-happy with the display of diff output, so I finally just knuckled down and switched over to difflib.  Given the following change to the 4.0 reference CQL file:

```
diff --git a/python/adelphi/tests/integration/resources/cql-schemas/4.0.0.cql b/python/adelphi/tests/integration/resources/cql-schemas/4.0.0.cql
index dbe0a97..c875e18 100644
--- a/python/adelphi/tests/integration/resources/cql-schemas/4.0-rc1.cql
+++ b/python/adelphi/tests/integration/resources/cql-schemas/4.0-rc1.cql
@@ -6,7 +6,7 @@ CREATE TYPE IF NOT EXISTS ks_0.udt_0 (
 );
 
 CREATE TYPE IF NOT EXISTS ks_0.udt_1 (
-    fld_2 frozen<udt_2>,
+    fld_22 frozen<udt_2>,
     fld_3 frozen<map<frozen<udt_2>, frozen<udt_2>>>,
     fld_4 frozen<list<frozen<udt_0>>>,
     fld_5 frozen<tuple<frozen<udt_2>, frozen<udt_0>>>
```

the output on test failure now looks like this:

```
Diff of generated file (4.0-rc1-stdout.out) against reference file (4.0-rc1.cql)                                                                                                  
--- /tmp/tmp6h5y4g6k/4.0-rc1-stdout.out                                                                                                                                           
+++ /work/git/adelphi/python/adelphi/tests/integration/resources/cql-schemas/4.0-rc1.cql                                                                                          
@@ -6,7 +6,7 @@                                                                                                                                                                   
);
                                      
CREATE TYPE IF NOT EXISTS ks_0.udt_1 (
-    fld_2 frozen<udt_2>,
+    fld_22 frozen<udt_2>,
fld_3 frozen<map<frozen<udt_2>, frozen<udt_2>>>,
fld_4 frozen<list<frozen<udt_0>>>,    
fld_5 frozen<tuple<frozen<udt_2>, frozen<udt_0>>>
```

absurdfarce:
Re: automated generation of tox.ini... this really should be in control of test-adelphi (since it needs to manipulate this file in order to set Cassandra version as well as several other options).  The intent is that whatever features we want to expose for tox.ini would migrate to CLI args of test-adelphi itself... something like this is done in this PR for keeping temp directories around after a run.  The big advantage of this approach is that it insulates users from having to know about tox, specifically about how to configure it... the app tries to hide as much of that as possible.

For the moment I've added a CLI flag to disable auto-generation of tox.ini, but going forward I'd much rather identify whatever functionality we want to include there and move it into the CLI rather than managing tox.ini outside of test-adelphi.

absurdfarce:
In addition to the changes addressing the specific comments in the earlier review I've also added a number of improvements/bug fixes discovered via testing.  Most notably I've attempted to enforce a deterministic order on keyspaces and tables within keyspace objects themselves (and, by extension, in their generated output).  We were seeing cases (especially with Python 2.7) where random ordering was messing with our difflib results in unpleasant ways.

python/adelphi/adelphi/export.py
```diff
@@ -95,7 +95,8 @@ def make_tuple(ks):
             if props['anonymize']:
                 anonymize_keyspace(ks)
             return KsTuple(ids[orig_name], ks)
-        return {t.ks_obj.name : t for t in [make_tuple(ks) for ks in keyspaces]}
+        tuples = sorted([make_tuple(ks) for ks in keyspaces], key=lambda ks: ks.ks_obj.name)
+        return OrderedDict([(t.ks_obj.name,t) for t in tuples])
```
absurdfarce, line 99: Keyspaces returned via export should be handled in an orderly way.  This is essential for legit compares of prior output to what's done by the current code, which in turn is an essential feature for our integration tests.

python/adelphi/adelphi/store.py
```diff
@@ -59,6 +60,10 @@ def build_keyspace_objects(keyspaces, metadata):
     """Build a list of cassandra.metadata.KeyspaceMetadata objects from a list of strings and a c.m.Metadata instance.  System keyspaces will be excluded."""
     all_keyspace_objs = [metadata.keyspaces[ks] for ks in keyspaces] if keyspaces is not None else metadata.keyspaces.values()
 
+    # Make sure tables are ordered (by table name)
+    for ks_obj in all_keyspace_objs:
+        ks_obj.tables = OrderedDict(sorted(ks_obj.tables.items(), key=lambda item: item[0]))
+
```
absurdfarce, line 66: A similar problem to what's discussed above: tables should also follow a coherent order.  Tests on Python 2.7 were seeing tables returned in an apparently random order, which in turn caused comparisons to a reference output to fail only because of the ordering of elements within the output.

python/adelphi/bin/test-adelphi
```diff
@@ -20,7 +20,7 @@ import tox
 
 
 # Default C* versions to include in all integration tests
-DEFAULT_CASSANDRA_VERSIONS = ["2.1.22", "2.2.19", "3.0.23", "3.11.10", "4.0-rc1"]
+DEFAULT_CASSANDRA_VERSIONS = ["2.1.22", "2.2.19", "3.0.23", "3.11.10", "4.0.0"]
```
absurdfarce, line 23: An update to 4.0.0... since I was already here :)

python/adelphi/tests/integration/__init__.py
```diff
@@ -62,6 +62,17 @@ def makeTempDirs(self):
...
+
+
+    def cleanUpTempDirs(self):
+        # TODO: Note that there's no easy way to access this from test-adelphi unless we modify the
+        # ini generation code... and I'm not completely sure that's worth it.  Might want to think
+        # about just deleting this outright... or making it a CLI option that can be easily accessed.
+        if "KEEP_LOGS" in os.environ:
+            log.info("KEEP_LOGS env var set, preserving logs/output at {}".format(self.dirs.basePath))
+        else:
+            shutil.rmtree(self.dirs.basePath)
```
absurdfarce, line 75: Switched to cleanup functions rather than an explicit tearDown() when trying to diagnose a different problem but in the end I decided to keep the change.  tearDown() does have some odd semantics; it's not called if setUp() doesn't succeed, for instance.  Rather than worry about it in the future it seemed saner to just convert to cleanup methods directly.

python/adelphi/tests/integration/test_cql.py
```diff
@@ -64,19 +73,43 @@ def buildSchema(self):
...
+        schemaPaths = glob.glob("{}/*/schema".format(outputDirPath))
+        self.assertGreater(len(schemaPaths), 0)
+        schemas = { extractKeyspaceName(p) : p for p in schemaPaths}
+        sortedKeyspaces = sorted(schemas.keys())
+
+        with open(allOutputPath, "w+") as allOutputFile:
+            cqlStr = "\n\n".join(open(schemas[ks]).read() for ks in sortedKeyspaces)
+            allOutputFile.write(cqlStr)
+
+        return allOutputPath
```
absurdfarce, line 112: Another problem observed in testing: the ordering of individual keyspace schemas returned by the glob above isn't deterministic so there's no guarantee we'll get the right ordering of keyspaces.  To fix this we extract the keyspace name from each individual keyspace schema and traverse them in order while generating the "all" schema.

PR #160, opened 2021-06-21
58 CI automation of test suite, take two

absurdfarce:


python/adelphi/adelphi/anonymize.py
```diff
@@ -15,7 +15,6 @@
 
 # Functions and constants related to the anonymization process
 from adelphi.store import get_standard_columns_from_table_metadata
-import re
```
absurdfarce, line 18: Randomly discovered unused import, presumably a holdover from previous changes

python/adelphi/adelphi/gh.py
```diff
@@ -22,7 +22,6 @@
 
 from github import Github
 
-logging.basicConfig(level=logging.INFO)
```
absurdfarce, line 25: This (and store.py below) should definitely not be setting up their own logging... they should be leveraging logs that are configured somewhere else.

python/adelphi/adelphi/store.py
```diff
@@ -39,20 +40,26 @@
...
-    if username is not None and password is not None:
-        auth_provider = PlainTextAuthProvider(username=username, password=password)
-    return auth_provider
+    """Instantiate auth provider if credentials have been provided"""
+    if username is None or password is None:
+        return None
+    return PlainTextAuthProvider(username=username, password=password)
 
 
+@retry
```
absurdfarce, line 49: Annotated to make this a tenacity-aware function but we don't provide global configs here since those are largely context-specific. Individual callers can provide args appropriate for their specific use; see the example in test-adelphi to see what this looks like.

python/adelphi/bin/test-adelphi
```diff
@@ -74,13 +109,13 @@ def runtests(cassandra, python, pytest):
             # exiting all the things
             try:
                 tox.cmdline(tox_args)
-            except SystemExit:
-                pass
+            except SystemExit as exc:
+                exitCodes.append(exc.code)
```
absurdfarce, line 113: The exit code that would be returned by tox lives in SystemExit.code. We need to trap this to determine how many of our tox "invocations" resulted in successful tests.

python/adelphi/test-requirements.txt
```diff
@@ -1,5 +1,2 @@
-click ~= 7.1
-cassandra-driver ~= 3.24
 docker ~= 4.4
-tenacity ~= 7.0
```
absurdfarce, line 4: Now that test-adelphi is a full-fledged part of the Python package (see setup.py change above) there isn't any need to manage it's dependencies via test requirements. This change was one of the significant arguments for making test-adelphi part of the package.

.github/workflows/adelphi-python.yaml
```diff
@@ -0,0 +1,29 @@
+name: adelphi Python
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
grighetto, line 7: Do we really need to specify the branch name?
We might change the name to trunk to match other repos and we'd have to change it here too.
Also, having it hardcoded won't it make it run less frequently, since all PRs are against master right now.

python/adelphi/adelphi/store.py
```diff
@@ -96,3 +101,28 @@ def set_replication_factor(selected_keyspaces, factor):
             log.debug("Replication for keyspace " + ks.name+ ": " + str(ks.replication_strategy))
             strategy = ks.replication_strategy
             strategy.replication_factor_info = factor
+
+
+def create_schema(session, schemaPath):
```
grighetto, line 106: Can we move this to the test module?

absurdfarce: We probably could, but I'd argue this function belongs in adelphi.store just as much as with_cluster() or with_local_cluster().  It seems like the bigger problem here is that it's poorly named.  This function exists to execute a collection of semicolon-terminated CQL statements contained in a file against a CQL instance somewhere.  We use it in the tests for setting up schemas but you can imagine any number of usages other than that within Adelphi.

I think the better answer would be to rename it to execute_cql() or something like that.  wdyt?

.github/workflows/adelphi-python.yaml
absurdfarce: Honestly I'm not sure if this field can be empty or not.  Most of the samples I found explicitly specified a target branch here but I didn't find any docs one way or the other.  That said, I guess I don't view this as a major burden.  If we decide to rename master at some point we know that'll be a disruptive change anyway; fixing this usage would just be one thing among many we'd have to address.  I also don't agree that all our PRs are off of master at the moment; it's certainly the case that most, even the overwhelming majority, are but there have been a few occasions where I've wanted review of a specific proposal for a feature branch already in-flight... and that leads to branches off of branches.  I readily agree it's not common but it does happen.

I'll do some checking to see whether a branch name is required here or not.

Issue #164, opened 2021-08-13
Internal registry not reachable with containerd


grighetto:
# Bug Report
### Description

Kubernetes [deprecated the Docker engine in version 1.20](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.20.md#deprecation) in favor of other engines such as containerd. GCP has also recently adopted containerd as its default engine for GKE clusters.
That led to some problems because the containerd daemon doesn't allow connections to insecure (non-TLS) registries, which is the case in Adelphi that runs an ephemeral registry within the cluster for storing images generated on the fly.

The difference in behavior comes from the fact that Docker checks if the registry address falls within the subnet and if so it connects with HTTP:

https://github.com/moby/moby/blob/e9b07a730ef02df793a28d07eb8ee3c39bba92d8/registry/config.go#L293

<pre><code>
	// Try CIDR notation only if addrs has any elements, i.e. if `host`'s IP could be determined.
	for _, addr := range addrs {
		for _, ipnet := range cidrs {
			// check if the addr falls in the subnet
			if (*net.IPNet)(ipnet).Contains(addr) {
				return true
			}
		}
	}
</code></pre>

Whereas containerd simply compares the address string to "localhost" or some loopback address "127.0.0.*" and if it doesn't match, it connects with HTTPS:

https://github.com/adisky/containerd/blob/8014d9fee0e2a758d34af977f1ea6a1b6932a098/pkg/cri/server/image_pull.go#L340
https://github.com/adisky/containerd/blob/8014d9fee0e2a758d34af977f1ea6a1b6932a098/pkg/cri/server/image_pull.go#L384

<pre><code>
			} else if isLocalHost(host) && u.Scheme == "http" {
				// Skipping TLS verification for localhost
				transport.TLSClientConfig.InsecureSkipVerify = true
			}
</code></pre>

We had already encountered this problem when testing with Kind (which also uses containerd), but for the test environment we had the option to disable the TLS check in the containerd configuration. We don't have that option though in a managed environment such as GKE, so we need a more robust solution now.

Adelphi is meant to be a standalone tool, with little to no external dependencies and capable of running in various environments (GKE, EKS, Rancher, Kind, etc), so the solution has to attain to those principles.

A few potential solutions include:

1. Setting up self-signed certificates for secure registry. **Problem:** Requires configuration in the worker nodes, which may vary with OS, Cloud vendor, etc.
[...]

PR #165, opened 2021-08-13
#164 Use a proxy sidecar to enable containerd <-> registry communication

grighetto:


helm/adelphi/templates/cassandra-clusters.yaml
```diff
@@ -78,7 +75,7 @@ spec:
           serverType: cassandra
           serverVersion: {{ .Values.target.version }}
           {{- if .Values.target.git_identifier }}
-          serverImage: {{`{{inputs.parameters.registry_ip}}`}}:30000/adelphi/management-api-for-apache-cassandra
+          serverImage: localhost:30000/adelphi/management-api-for-apache-cassandra
```
grighetto, line 78: containerd only allows connections to insecure registries running on localhost, so now we have a proxy running locally on port 30000 that in turn points to the real registry running at the `docker-registry` service.
So we don't need the `registry_ip` parameter anymore, which in one way is also a simplification.

helm/adelphi/templates/cassandra-image-builder.yaml
```diff
@@ -10,20 +10,33 @@ spec:
   - name: cassandra-image-template
     inputs:
       parameters:
-      - name: registry_ip
       - name: git_identifier
     container:
-      image: gcr.io/kaniko-project/executor:v1.0.0
+      image: gcr.io/kaniko-project/executor:v1.6.0-debug
```
grighetto, line 15: The general-purpose Kaniko image extends Docker's scratch image, which is a really bare system with no networking utilities and no way to install them. Fortunately, the Kaniko project has this `-debug` version that extends busybox which at least has `wget` installed. We need wget to check if the sidecar proxy is already running  before running the Kaniko executor, otherwise it will fail immediately.

helm/adelphi/templates/cassandra-image-builder.yaml
```diff
@@ -10,20 +10,33 @@ spec:
...
+      command:
+        - sh
+        - -c
       args:
-      - --dockerfile=/dockerfile/Dockerfile
-      - --context=dir:///workspace
-      - {{`--destination={{inputs.parameters.registry_ip}}:30000/adelphi/cassandra`}}
-      - --insecure=true
+        - until `wget --spider http://localhost:30000`; do echo sleep && sleep 1; done;
```
grighetto, line 20: The `--spider` flag will perform a HEAD request instead of a GET just to check if the proxy is up and running, it won't care about the respond body.

helm/adelphi/templates/cassandra-image-builder.yaml
```diff
@@ -44,9 +57,9 @@ spec:
         - sh
         - -c
       args:
-        - wget -O - https://downloads.apache.org/ant/binaries/apache-ant-1.9.15-bin.tar.gz | tar -zxvf -;
-          chmod +x /apache-ant-1.9.15/bin/ant;
-          ln -s /apache-ant-1.9.15/bin/ant /usr/bin/ant;
+        - wget -O - https://downloads.apache.org/ant/binaries/apache-ant-1.10.11-bin.tar.gz | tar -zxvf -;
+          chmod +x /apache-ant-1.10.11/bin/ant;
+          ln -s /apache-ant-1.10.11/bin/ant /usr/bin/ant;
```
grighetto, line 62: That Apache downloads page doesn't keep the list of all Ant releases, it seems it only keeps the X most recent ones, so this pod was failing because it couldn't download that version anymore.
We should probably find an suitable image in Dockerhub or publish our own under the adelphitools org.

helm/adelphi/templates/cassandra-image-builder.yaml
```diff
@@ -67,3 +80,6 @@ spec:
     - name: dockerfile
       configMap:
         name: cassandra-dockerfile-configmap
+    - name: proxy-config
+      configMap:
+        name: registry-proxy-configmap
```
grighetto, line 85: This is where the nginx proxy config comes from.

helm/adelphi/templates/cassandra-mgmt-dockerfile-configmap.yaml
```diff
@@ -19,7 +19,7 @@ data:
     COPY management-api-server/pom.xml ./management-api-server/pom.xml
     # this duplicates work done in the next steps, but this should provide
     # a solid cache layer that only gets reset on pom.xml changes
-    RUN mvn -q -ff -T 1C install && rm -rf target
+    RUN mvn -q -ff -T 1C install -DskipTests && rm -rf target
```
grighetto, line 22: This speeds up the execution for me (we shouldn't be re-testing the MGMT API in our end either).

helm/adelphi/templates/cassandra-mgmt-image-builder.yaml
```diff
@@ -33,7 +43,7 @@ spec:
         - -c
       args:
         - git clone https://github.com/datastax/management-api-for-apache-cassandra .;
-          git checkout v0.1.25;
+          git checkout v0.1.27;
```
grighetto, line 46: Upgraded to the most recent version of the MGMT API

helm/adelphi/templates/cassandra-mgmt-dockerfile-configmap.yaml
```diff
@@ -34,7 +34,7 @@ data:
         if test ! -e datastax-mcac-agent-${METRICS_COLLECTOR_VERSION}.tar.gz; then curl -L -O "https://github.com/datastax/metric-collector-for-apache-cassandra/releases/download/v${METRICS_COLLECTOR_VERSION}/datastax-mcac-agent-${METRICS_COLLECTOR_VERSION}.tar.gz"; fi && \
         tar --directory ${MCAC_PATH} --strip-components 1 --gzip --extract --file datastax-mcac-agent-${METRICS_COLLECTOR_VERSION}.tar.gz
 
-    FROM ${REGISTRY}:30000/adelphi/cassandra
+    FROM localhost:30000/adelphi/cassandra
```
grighetto, line 37: Use the proxy

helm/adelphi/templates/registry-proxy-configmap.yaml
```diff
@@ -0,0 +1,16 @@
...
+  namespace: {{ .Values.namespace }}
+data:
+  default.conf: |
+    client_max_body_size 1000M;
+    server {
+      listen        30000;
+      location / {
+        proxy_set_header   X-Forwarded-For $remote_addr;
+        proxy_set_header   Host $http_host;
+        proxy_pass         "http://docker-registry.{{ .Values.namespace }}:30000/";
```
grighetto, line 14: Redirect requests to the actual registry.

helm/adelphi/templates/registry-proxy-configmap.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: registry-proxy-configmap
+  namespace: {{ .Values.namespace }}
+data:
+  default.conf: |
+    client_max_body_size 1000M;
```
grighetto, line 8: Increased max request size such that it can take the generated docker images.

jdonenine (approved):
Pretty unfortunate that we have to jump through these hoops because of the containerd change, but the solution is a good creative one I think.

Probably just one question, but overall, looks good 👍 

Should we try to constrain the resources allocated to the sidecar container?  Running that nginx process shouldn't be heavyweight, but we could forcibly constrain it as necessary too.

grighetto:
> Should we try to constrain the resources allocated to the sidecar container

@jdonenine Good point. I just set the resource limits for nginx.
I'll test the limits in Github Actions in the next PR and tweak it if necessary.

Issue #38, opened 2020-10-27
Schedule Smoke Testing Integration


grighetto:
The goal of this issue is to include some smoke tests that verify the basic Adelphi features.
We want to catch things like incompatibilities of running  Adelphi or the Cass-Operator on the latest version of k3d.

We can also leverage this work to produce a routine/scheduled cadence of tests against C* trunk to externally verify its continued reliability over time.

┆Issue is synchronized with this [Jira Task](https://datastax-oss.atlassian.net/browse/AD-88) by [Unito](https://www.unito.io)
┆Issue Number: AD-88

jdonenine:
Which direction should we go:
* GKE hosted resources (simpler but more costly)
* GH Actions (cheaper but more complicated and perhaps not even possible)

PR #166, opened 2021-08-14
#38 Workflow execution in GitHub actions

grighetto:
TO BE REVIEWED AFTER https://github.com/datastax/adelphi/pull/165

Example of successful workflow execution can be found here https://github.com/datastax/adelphi/actions/runs/1129534522


.github/workflows/cluster-integration-tests-kind.yaml
```diff
@@ -0,0 +1,41 @@
+name: Cluster-based Integration Tests (KIND)
+on: [push, pull_request]
+jobs:
+  integration-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          token: ${{ github.token }}
```
grighetto, line 10: This allows for testing locally with https://github.com/nektos/act

.github/workflows/cluster-integration-tests-kind.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Create Kind Cluster
+        uses: helm/kind-action@v1.1.0
+        with: 
+          version: v0.11.1
+          node_image: kindest/node:v1.21.1
+          cluster_name: adelphi-integration-tests
+          config: helm/adelphi/tests/kind-config-3-workers.yaml
+      - name: Install Chart and Run Workflow
+        run: helm install adelphi helm/adelphi -n cass-operator -f helm/adelphi/tests/kind-values-trunk.yaml --debug --wait || true
+      - name: Install Chart and Run Workflow (2nd attempt in case of timeout)
```
grighetto, line 31: In a constrained environment, it might take over a minute to install the two CRDs we currently have. If it times out, it won't complete the Adelphi installation and we have to try again, so this is what this step does.
On the other hand, it the first installation succeeded, it will simply ignore the second attempt.

In future If we can contribute a patch to helm such that the timeout is configurable, we can remove this step.

.github/workflows/cluster-integration-tests-kind.yaml
```diff
@@ -0,0 +1,41 @@
...
+        run: helm install adelphi helm/adelphi -n cass-operator -f helm/adelphi/tests/kind-values-trunk.yaml --debug || true
+      - name: Wait for Workflow Server ready status
+        run: |
+          sleep 30
```
grighetto, line 35: `kubectl wait` will fail right away if it can't even find the pod it's supposed to wait on (again, might happen in constrained environments), so give it some time to settle.

.github/workflows/cluster-integration-tests-kind.yaml
```diff
@@ -0,0 +1,41 @@
...
+          kubectl get all -A
+          ./argo-linux-amd64 get -n cass-operator @latest
```
grighetto, line 37: Print some debugging information so we can check for any start up failures.

helm/adelphi/kind-values.yaml
```diff
@@ -0,0 +1,5 @@
+############################ 
+# Config overrides for KIND 
+############################
+
+storageClassName: standard
```
grighetto, line 5: Overrides the `local-path` storage class that's the default value in Adelphi (for k3d).

helm/adelphi/tests/kind-config-3-workers.yaml
```diff
@@ -0,0 +1,8 @@
+# 3-worker cluster config
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+- role: worker
+- role: worker
+- role: worker
```
grighetto, line 8: This is the Kind cluster configuration for the integration tests, with 3 worker nodes.

helm/adelphi/kind-values.yaml
grighetto: This is not used by the tests, it's just a sample config for users running on Kind (similar to what we have for GKE).

helm/adelphi/tests/kind-values-trunk.yaml
```diff
@@ -0,0 +1,15 @@
+############################ 
+# Config overrides for KIND
+# Used in integration tests
+# Runs against C* trunk 
+############################
+
+storageClassName: standard
+
+clusterSize: 1
```
grighetto, line 9: This is the values file used by the integration tests for running on Kind in GithubActions.
We run with a single node on each cluster to avoid hitting the env limits.

helm/adelphi/tests/kind-values-trunk.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+target:
+  dc: dc2
+  clusterName: target
+  version: "4.0.0"
+  git_identifier: trunk
```
grighetto, line 15: The tests in GithubActions will run against C* trunk.

.github/workflows/cluster-integration-tests-kind.yaml
```diff
@@ -0,0 +1,41 @@
...
+          kubectl wait pod --for=condition=Ready -l app=argo-server -n cass-operator --timeout=10m
+      - name: Argo Wait for Completion
+        run: |
+          ./argo-linux-amd64 watch -n cass-operator @latest
```
grighetto, line 41: `argo watch` is not ideal here because it will spam the logs, but I wouldn't like to hold this back because of it, we can improve it later.
Ideally, we should propose/contribute a refresh rate argument for the Argo project.

.github/workflows/cluster-integration-tests-kind.yaml
jdonenine: Here you're talking a timeout simply while installing the chart?  Right?  I think that's the case, just want to confirm.  

The timeout they do have on the cli tool won't cover the case you're talking about here? With the initial installation of the CRDs?

```
--timeout duration             time to wait for any individual Kubernetes operation (like Jobs for hooks) (default 5m0s)
```

.github/workflows/cluster-integration-tests-kind.yaml
jdonenine: What will happen if the chart installation itself fails?  Will it kick out here as a overall failure of the workflow?

.github/workflows/cluster-integration-tests-kind.yaml
jdonenine: One one hand, it might be verbose, but on the other we've seen many times in K8ssandra how difficult it can be to debug issues that we see in the GHA workflows, so having the logging will be quite useful when things go wrong.  As long as it's not so much that we hit log size limits or anything like that, I don't think it's a huge problem and in some ways desirable.

jdonenine (commented):
This is a great add @grighetto great to get a full integration test in there that we can use as we move forward.  Looks like the test run took around 20 minutes, so pretty reasonable speed wise to use on PRs.

I added a few comments to discuss, thanks!

.github/workflows/cluster-integration-tests-kind.yaml
grighetto: @jdonenine I confirmed my suspicion that the `--timeout` flag only applies to the regular resources, not to the CRDs.
Installing the chart with:

```
helm install adelphi helm/adelphi -n cass-operator -f helm/adelphi/tests/kind-values-trunk.yaml --debug --wait --timeout=10m
```

Prints the following for the CRDs section:

```
wait.go:47: [debug] beginning wait for 29 resources with timeout of 1m0s
```

And the following for the other resources:

```
wait.go:47: [debug] beginning wait for 26 resources with timeout of 10m0s
```

It seems the timeout for the CRDs is static at 1min. We can always revisit this in the future when it becomes tunable.

.github/workflows/cluster-integration-tests-kind.yaml
grighetto: `argo get` always exits with 0, so it won't fail exactly here, but it will fail in later steps (when it tries to connect to the results server).

.github/workflows/cluster-integration-tests-kind.yaml
grighetto: Alright, sounds good. So far we haven't hit any limits and it actually came in handy a couple of times while debugging during development.

.github/workflows/cluster-integration-tests-kind.yaml
jdonenine: Well that's unfortunate 😔 

Think we should file an enhancement request with helm to allow for the additional timeouts to be set related to CRD install?

jdonenine (approved):
Took a look over the commits since my last review, looks good.  I added just one comment about the helm timeouts, but that's now out of band with this here, so good to go.  Thanks!

Issue #132, opened 2021-05-16
Allow for using different schemas for the source and target clusters


grighetto:
# Feature Request
### Description

By allowing different schemas in each cluster, we'd achieve two goals:

* While schemas are not automatically migrated, we'd allow for manually fixing any incompatibilities such that they can be used with Adelphi;
* Comparing variations of the same schema with source and target being the same C* version (not the main goal of the project, but seems a nice addition).

### Additional context

┆Issue is synchronized with this [Jira Task](https://datastax-oss.atlassian.net/browse/AD-24) by [Unito](https://www.unito.io)
┆Issue Number: AD-24

PR #169, opened 2021-08-19
#132 Allow for using different CQL schemas per cluster.

grighetto:


jdonenine (approved):
We discussed the need for this on Slack, so all good on that front.  It's an additional option for situations where because of difference in the clusters the same exact schema is not usable on each.

Issue #129, opened 2021-05-14
NB template - allow setting `cycles` as template param


grighetto:
# Bug Report

### Description

Allow for setting main and ramp-up cycles again as template vars. 
Initially reported in https://github.com/datastax/adelphi/pull/114/files#r606063843 and I just stumbled across it while testing on GKE, it took me a moment to understand why the cycle params were not becoming effective.

### Environment

 

**Python Version:**

<pre><code>

</code></pre>

**pip Version:**

<pre><code>

</code></pre>

**Anonymizer Package Version:**

<pre><code>

</code></pre>

**Kubernetes Version:**

<pre><code>

</code></pre>

**Kubernetes Cluster Type:**

**Helm Version:**

<pre><code>

</code></pre>

**Helm Charts:**

<pre><code>

</code></pre>

**Argo Version:**

<pre><code>

</code></pre>

**Other relevant information:**

PR #171, opened 2021-08-19
#129 Make `rampup-cycles` and `main-cycles` be NB template vars again.

grighetto:
This will require a new pypi release to become effective.

### DIFF
--- a/helm/adelphi/templates/nosqlbench-jobs.yaml
+++ b/helm/adelphi/templates/nosqlbench-jobs.yaml
@@ -17,6 +17,8 @@ spec:
         "/workspace/nosqlbench_source.yaml",
         "source",
         "scenarioname=source",
+        "rampup-cycles={{ .Values.nosqlbench_cycles }}",
+        "main-cycles={{ .Values.nosqlbench_cycles }}",
         "hosts={{ .Values.source.clusterName }}-{{ .Values.source.dc }}-service",
         "username={{ .Values.source.clusterName }}-superuser",
         "passfile=/secret/password"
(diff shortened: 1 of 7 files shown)


helm/adelphi/templates/nosqlbench-jobs.yaml
```diff
@@ -39,7 +41,7 @@ spec:
       command: ["sh", "-c", "-e"]
       args:
         - pip install adelphi;
-          adelphi --hosts="{{ .Values.source.clusterName }}-{{ .Values.source.dc }}-service" --username="{{ .Values.source.clusterName }}-superuser" --password="`cat /secret/password`" --keyspaces="{{ .Values.gemini_schema_keyspace }}" --no-anonymize export-nb --rampup-cycles="{{ .Values.nosqlbench_cycles }}" --main-cycles="{{ .Values.nosqlbench_cycles }}" > /workspace/nosqlbench_source.yaml;
+          adelphi --hosts="{{ .Values.source.clusterName }}-{{ .Values.source.dc }}-service" --username="{{ .Values.source.clusterName }}-superuser" --password="`cat /secret/password`" --keyspaces="{{ .Values.gemini_schema_keyspace }}" --no-anonymize export-nb > /workspace/nosqlbench_source.yaml;
```
jdonenine, line 44: So, I get the change here, but I think I probably need to understand a bit more about what the python tooling spits out for the template.

Maybe and first question... why were these two parameters provided for within the python tooling to begin with?  And why is not having them there the "right" way to do this?

grighetto: The goal here is to separate execution parameters from the template generation parameters.
The number of cycles is something that matters only at runtime, so once you have the workload YAML generated, you can tweak the number of cycles as you wish without generating the YAML all over again (which requires a connection to the cluster).

Another benefit is keeping the adelphi cli slim and relying on the existing variable replacement feature in NoSQLBench instead.

